PR #1, opened 2015-01-12
add packer default build

phrawzty:
@rhelmer `r?`

### DIFF
--- a/packer/puppet/modules/socorro/files/etc_elasticsearch/elasticsearch.yml
+++ b/packer/puppet/modules/socorro/files/etc_elasticsearch/elasticsearch.yml
@@ -1,0 +1,3 @@
+---
+index.number_of_shards: 1
+index.number_of_replicas: 0
--- a/packer/puppet/modules/socorro/files/var_lib_pgsql_9.3_data/pg_hba.conf
+++ b/packer/puppet/modules/socorro/files/var_lib_pgsql_9.3_data/pg_hba.conf
@@ -1,0 +1,89 @@
+# PostgreSQL Client Authentication Configuration File
+# ===================================================
+#
+# Refer to the "Client Authentication" section in the PostgreSQL
+# documentation for a complete description of this file.  A short
+# synopsis follows.
+#
+# This file controls: which hosts are allowed to connect, how clients
+# are authenticated, which PostgreSQL user names they can use, which
+# databases they can access.  Records take one of these forms:
+#
+# local      DATABASE  USER  METHOD  [OPTIONS]
+# host       DATABASE  USER  ADDRESS  METHOD  [OPTIONS]
+# hostssl    DATABASE  USER  ADDRESS  METHOD  [OPTIONS]
+# hostnossl  DATABASE  USER  ADDRESS  METHOD  [OPTIONS]
+#
+# (The uppercase items must be replaced by actual values.)
+#
+# The first field is the connection type: "local" is a Unix-domain
+# socket, "host" is either a plain or SSL-encrypted TCP/IP socket,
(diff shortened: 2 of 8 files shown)


packer/puppet/modules/socorro/files/var_lib_pgsql_9.3_data/pg_hba.conf
```diff
@@ -0,0 +1,89 @@
...
+# configuration parameter, or via the -i or -h command line switches.
+
+
+
+# TYPE  DATABASE        USER            ADDRESS                 METHOD
+
+# "local" is for Unix domain socket connections only
+local   all             all                                     peer
+# IPv4 local connections:
+host    all             all             127.0.0.1/32            trust
```
rhelmer: This is OK for dev (vagrant) but we should require passwords for production, e.g. md5 http://www.postgresql.org/docs/current/static/auth-methods.html

packer/puppet/modules/socorro/files/var_lib_pgsql_9.3_data/pg_hba.conf
```diff
@@ -0,0 +1,89 @@
...
+# IPv6 local connections:
+host    all             all             ::1/128                 trust
```
rhelmer: Same for as ipv4 above, ```s/trust/md5/```

packer/puppet/modules/socorro/manifests/init.pp
```diff
@@ -0,0 +1,111 @@
...
+  service {
+    'httpd':
+      ensure  => stopped,
+      enable  => false,
+      require => Package['httpd'];
+
+    'memcached':
+      ensure  => stopped,
+      enable  => false,
+      require => Package['memcached'];
```
rhelmer: Could we remove this and use elasticache instead?

packer/puppet/modules/socorro/files/etc_elasticsearch/elasticsearch.yml
```diff
@@ -0,0 +1,3 @@
+---
+index.number_of_shards: 1
+index.number_of_replicas: 0
```
rhelmer: Presumably this isn't going to be good enough in prod, how are we going to handle pushing config changes?

rhelmer:
lgtm overall, a few production-y concerns.

packer/puppet/modules/socorro/manifests/init.pp
phrawzty: I wasn't sure if we had settled on using an Amazon service or not.  I can remove it if we're for sure headed in that direction.

packer/puppet/modules/socorro/files/etc_elasticsearch/elasticsearch.yml
phrawzty: I debated including *any* config files, to be honest.  Now that I'm looking at it again, I'm thinking that they should all be removed entirely since - in principle - they will be generated at instance provision / deploy time.  What do you think?

packer/puppet/modules/socorro/manifests/init.pp
rhelmer: Looks dead simple to support in Django (the only service that needs memcached), I think we should go for it.

packer/puppet/modules/socorro/files/etc_elasticsearch/elasticsearch.yml
rhelmer: I think having usable defaults that get overridden at deploy time would be a good compromise - that would mean that we're producing public AMIs that can actually work out of the box, but need a bunch of new configs dropped in if you want to run a custom or distributed system.

rhelmer:
r+

PR #5, opened 2015-01-29
centos7 compatible

phrawzty:
There are some differences between the cloud images for CentOS 6 and 7. This PR addresses those issues. As an added bonus: Elasticsearch 1.4, too. :smile:

@rhelmer `r?`

### DIFF
--- a/packer/puppet/modules/socorro/manifests/init.pp
+++ b/packer/puppet/modules/socorro/manifests/init.pp
@@ -28,7 +28,7 @@ class socorro::generic {
 
   yumrepo {
     'elasticsearch':
-      baseurl => 'http://packages.elasticsearch.org/elasticsearch/0.90/centos',
+      baseurl => 'http://packages.elasticsearch.org/elasticsearch/1.4/centos',
       gpgkey  => 'https://packages.elasticsearch.org/GPG-KEY-elasticsearch';
     'PGDG':
       baseurl => 'http://yum.postgresql.org/9.3/redhat/rhel-$releasever-$basearch',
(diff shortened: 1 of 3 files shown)


packer/puppet/modules/socorro/manifests/init.pp
```diff
@@ -53,7 +53,6 @@
 
   package {
     [
-      'daemonize',
```
rhelmer: I noticed this was gone which breaks our init scripts (crashmover and processor) - fine I guess since we want to stop using these anyway!

packer/puppet/modules/socorro/manifests/init.pp
```diff
@@ -28,7 +28,7 @@
 
   yumrepo {
     'elasticsearch':
-      baseurl => 'http://packages.elasticsearch.org/elasticsearch/0.90/centos',
+      baseurl => 'http://packages.elasticsearch.org/elasticsearch/1.4/centos',
```
rhelmer: Oh are we ready for this? Cool!

rhelmer:
r+

packer/puppet/modules/socorro/manifests/init.pp
phrawzty: At first glance this is the only *breaking* change caused by a move to CentOS 7 - and as you said, it will force us to deal with cruft, which is a good thing.

packer/puppet/modules/socorro/manifests/init.pp
phrawzty: The code passes tests and the feedback from the community has been positive - so yeah, why the hell not? :wine_glass:

AdrianGaudebert:
Elasticsearch 1.4.

AdrianGaudebert:
Though we should probably change the defaults to use the new ``es`` package instead of ``elasticsearch``, because the later is going to break with ES 1.4.

PR #8, opened 2015-02-09
initial commit for terraform

phrawzty:
This is a working model for a Socorro deployment to AWS using Terraform.

@rhelmer `r?`

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -16,6 +16,6 @@ process are still being worked on.
 ## Details
 
 Performing a `terraform apply` on this repo will result in a small amount of
-infrastructure being spun up in the `eu-west-1` AWS zone.  The AWS access and
+infrastructure being spun up in the `us-west-2` AWS zone.  The AWS access and
 secret keys need to be supplied either as environment variables, or via
 the `terraform.tfvars` file.
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -1,0 +1,21 @@
+# Terraform
+
+[Terraform](https://www.terraform.io) is a tool for building and maintaining
+virtual infrastructure in a variety of environments.  We use it for managing
+various elements of Socorro's cloud-based deployment.
+
+## Socorro model
+
+The AMI used for spinning up new instances is built using Packer (also in this
+repo).  This AMI comes pre-installed with all of the bits and pieces required
+to run pretty much every aspect of Socorro; however, by default, all of the
+possible services are disabled.  There are a number of *roles* which determine
+which aspects of the default AMI are activated - the details of this activation
+process are still being worked on.
+
+## Details
+
+Performing a `terraform apply` on this repo will result in a small amount of
+infrastructure being spun up in the `eu-west-1` AWS zone.  The AWS access and
+secret keys need to be supplied either as environment variables, or via
+the `terraform.tfvars` file.
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -89,6 +89,19 @@ resource "aws_security_group" "elb_to_webheads__http" {
     }
 }
 
+resource "aws_security_group" "internet_to_snowflakes__http" {
+    name = "internet_to_snowflakes__http"
+    description = "Allow HTTP access to some oddball nodes."
+    ingress {
+        from_port = 80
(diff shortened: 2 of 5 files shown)


terraform/README.md
```diff
@@ -0,0 +1,21 @@
...
+repo).  This AMI comes pre-installed with all of the bits and pieces required
+to run pretty much every aspect of Socorro; however, by default, all of the
+possible services are disabled.  There are a number of *roles* which determine
+which aspects of the default AMI are activated - the details of this activation
+process are still being worked on.
+
+## Details
+
+Performing a `terraform apply` on this repo will result in a small amount of
+infrastructure being spun up in the `eu-west-1` AWS zone.  The AWS access and
```
rhelmer: Our S3 buckets are in us-west-2 - should we change this now?

phrawzty: Via a combination of `lookup()` and the selector-based variable assignments in `variables.tf`, we can set up any number of zones.  It would make sense for the basic default zone to be `us-west-2`; however, I haven't set up keys or the default image in that zone yet.  Once that zone is provisioned we can (and should) change the default here.

terraform/main.tf
```diff
@@ -0,0 +1,257 @@
...
+    availability_zones = [
+        "${aws_instance.webheads.*.availability_zone}"
+    ]
+    listener {
+        instance_port = 80
+        instance_protocol = "http"
+        lb_port = 80
+        lb_protocol = "http"
+    }
+    /* Requires SSLCertificateId
```
rhelmer: what's the plan for this?

terraform/main.tf
```diff
@@ -0,0 +1,257 @@
...
+
+resource "aws_instance" "collectors" {
+    ami = "${lookup(var.base_ami, var.region)}"
+    instance_type = "t2.micro"
+    key_name = "${lookup(var.ssh_key_name, var.region)}"
+    count = 1
+    security_groups = [
+        "${aws_security_group.elb_to_webheads__http.name}",
+        "${aws_security_group.internet_to_any__ssh.name}",
+		"${aws_security_group.private_to_private__any.name}"
```
rhelmer: formatting (tabs instead of spaces)

terraform/main.tf
```diff
@@ -0,0 +1,257 @@
...
+}
+
+resource "aws_instance" "admin_host" {
+    ami = "${lookup(var.base_ami, var.region)}"
+    instance_type = "t2.micro"
+    key_name = "${lookup(var.ssh_key_name, var.region)}"
+    security_groups = [
+        "${aws_security_group.internet_to_any__ssh.name}",
+		"${aws_security_group.private_to_private__any.name}"
+    ]
```
rhelmer: We are picking up a few more nodes as part of the move, if you want to add these now:

* crash-analysis
* symbolapi

terraform/main.tf
phrawzty: Looks like an errant search and replace in vim was the culprit. :baby:

terraform/main.tf
phrawzty: Do these need ELBs to front them ?

rhelmer: No I think they can just be single standalone nodes (that's what they are now), they both need to be public-facing though.

terraform/main.tf
phrawzty: We need to integrate the certificates into AWS.  Once the certs are available they can be applied in the ELB context, for example:
```ini
    listener {
        ssl_certificate_id = "arn:aws:iam::123456789012:server-certificate/certName"
    }
```
That's simple enough - it's the AWS component that I'm less sure about, heh.

terraform/README.md
phrawzty: Addressed in 46f3998.

rhelmer: Does the README need to be updated? ```variables.tf``` lists ```default = "us-west-2"``` now

phrawzty:
@rhelmer renewed `r?`

terraform/README.md
phrawzty: Good catch!

rhelmer:
r+

PR #11, opened 2015-02-12
fix bug 1132398; add buildbox profile & support for alt profiles in general

phrawzty:
This PR introduces the idea of alternative Packer profiles, and adds a `buildbox` profile into the mix.

This is more or less just a bunch of Puppet stuff.  There are any number of ways to deal with this - including layers of templates, use of Hiera, and so forth.  I opted to keep it simple, which is fine as long as we don't need to extend this to, like dozens of profiles. :smile: 

Anyway, what we need to do is determine what needs to be installed on the buildbox (I just put some of the stuff from the Vagrant box in there as filler, for now).  This is something we'll need to work on together, @rhelmer . :grapes:

### DIFF
--- a/packer/puppet/manifests/default.pp
+++ b/packer/puppet/manifests/default.pp
@@ -3,5 +3,12 @@ Exec {
 }
 
 node default {
-  include socorro::generic
+  case $::packer_profile {
+    'base': { include socorro::base }
+    'buildbox': { include socorro::buildbox }
+    default: {
+      err("'${::packer_profile}' is not a valid Packer profile label.")
+      fail('Invalid packer_profile.')
+    }
+  }
 }
--- a/packer/puppet/modules/socorro/manifests/base.pp
+++ b/packer/puppet/modules/socorro/manifests/base.pp
@@ -1,0 +1,109 @@
+# Set up basic Socorro requirements.
+class socorro::base {
+
+  include socorro
+
+  service {
+    'httpd':
+      ensure  => stopped,
+      enable  => false,
+      require => Package['httpd'];
+
+    'postgresql-9.3':
+      ensure  => stopped,
+      enable  => false,
+      require => [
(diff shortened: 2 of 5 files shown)


rhelmer:
This looks fine in principle... the packages in the ```socorro::buildbox``` class lgtm.

Really all this node needs to be able to do is run ```make package BUILD_TYPE=rpm``` and then put the RPM in a public S3 bucket, and terraform can use it to provision the actual Socorro nodes.

It would be pretty slick if we used the travis deploy support to trigger this maybe host a git repo a la http://docs.travis-ci.com/user/deployment/custom/ 

I know @lonnen has done similar setups for deployment - any suggestions?

lonnen:
Static files are pretty easy. In the travis manifest you can specify a directory to put into an S3 bucket, so dump your RPMs out to a dir and they can be uploaded directly or s3cmd can do it for us (I think. Don't have my noted on me about the exact utility)

phrawzty:
> It would be pretty slick if we used the travis deploy support to trigger this maybe host a git repo a la http://docs.travis-ci.com/user/deployment/custom/

I'm all for letting Travis do the work here; however, Travis doesn't have the same arch as we do, which means that any compiled binaries will likely not be portable (unless we do some serious dancing).  I'm thinking specifically of Stackwalker in this case, and while I realise that this isn't part of the Socorro RPM, but it *is* something that we need to build and support on our architecture.  Other considerations include, well, anything that gets compiled when the Virtualenv is built, now or in the future for as long as we're not using exactly the same arch as Travis.

rhelmer:
@phrawzty sorry I was not clear here - what I am proposing is that travis runs tests only, and notifies the build server when they pass via the "deploy" mechanism.

Our build server produces build and uploads the artifact(s) to S3.

phrawzty:
Sounds good. We'll roll with this model for now.

PR #12, opened 2015-02-13
fix bug 1132180; add environment prefix

phrawzty:
This also adds tags to many of the resources, the net effect being that those resources are easier to identify via the console (which is handy for testing).  Speaking of testing, note the `del_on_term` variable, too.

@rhelmer `r?` :cactus:

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -6,7 +6,7 @@ provider "aws" {
 
 # This is potentially dangerous; may require review.
 resource "aws_security_group" "private_to_private__any" {
-    name = "private_to_private__any"
+    name = "${var.environment}__private_to_private__any"
     description = "Allow all private traffic."
     ingress {
         from_port = 0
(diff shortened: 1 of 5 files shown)


terraform/main.tf
```diff
@@ -149,22 +164,22 @@ resource "aws_elb" "elb_for_webheads" {
...
-        "${aws_instance.webheads.*.id}"
+        "${aws_instance.webhead.*.id}"
     ]
     security_groups = [
         "${aws_security_group.internet_to_elb__http.id}"
     ]
 }
 
-resource "aws_instance" "webheads" {
+resource "aws_instance" "webhead" {
```
rhelmer: why don't the ```aws_instance``` resources have the ```${var.environment}``` prefix, because they are just random EC2 instances? Is there a way we can set the "notes" field that appears in the console?

rhelmer:
lgtm but I wonder why all the EBS volumes were added in this PR?

terraform/main.tf
phrawzty: They *do* have the prefix: look down to lines 204 thru 207 (in this case).  All of the `aws_instance` resources are configured to have both `Name` and `Environment` tags, the former being applied in the Name column in the AWS web interface, and the latter being a handy way to sort things via both the web interface and CLI tools.

As an aside, the `resource` declaration is an internal label that is used to build the Terraform resource graph; variables are not permitted here in any case.

phrawzty:
Right now, all of these are `t2.micro` instance types, and as such, only support EBS volumes (they do not have instance storage).  Since this is the only option, those volumes do not normally need to be declared explicitly, which is fine; however, if one wants to *activate* the `delete_on_termination` option, one must declare the resource explicitly, hence their appearance here.

See also line 20 of `README.md` in this PR:
> For testing purposes, it is useful to enable `del_on_term`, which will remove EBS volumes once their associated instances are terminated.

PR #13, opened 2015-02-19
more deps for building packages

phrawzty:
Required for building things as a user *other* than `root`, heh.

@rhelmer `r?`


rhelmer:
r+

PR #14, opened 2015-02-20
update basebox ami ID & remove refs to EU

phrawzty:
This is going to suck less once we move to dynamic storage.

@rhelmer `r?`


rhelmer:
r+

PR #15, opened 2015-02-25
Bug1127532 symbolapi production

rhelmer:
This PR clears out everything that isn't production-ready (all that stuff is on the new ```stage``` branch instead)

Some open questions:

* how do we deal with DNS? We want this to be http://symbolapi.mozilla.org
* should there be an output variable for this (the public IP?)
* would it be valuable to use an ELB even for a single node like this? Would that e.g. enable us to automatically bring up a new VM if this one failed?

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -27,3 +27,8 @@ keys in `terraform.tfvars` and specifying the rest at runtime:
 terraform plan -var 'environment=foo' -var 'del_on_term=true' -out=foo.plan
 terraform apply foo.plan
 ```
+
+## Making changes live in staging and production
+
+Anything landed on master will go to staging and be considered eligible for
+pushing to production. Tags will be pushed to production.
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -53,54 +53,6 @@ resource "aws_security_group" "internet_to_any__ssh" {
     }
 }
 
-resource "aws_security_group" "internet_to_elb__http" {
-    name = "${var.environment}__internet_to_elb__http"
-    description = "Allow incoming traffic from Internet to HTTP(S) on ELBs."
-    ingress {
-        from_port = 80
-        to_port = 80
-        protocol = "tcp"
-        cidr_blocks = [
-            "0.0.0.0/0"
-        ]
-    }
-    ingress {
-        from_port = 443
-        to_port = 443
-        protocol = "tcp"
-        cidr_blocks = [
-            "0.0.0.0/0"
-        ]
-    }
-    tags {
-        Environment = "${var.environment}"
-    }
-}
-
-resource "aws_security_group" "elb_to_webhead__http" {
-    name = "${var.environment}__elb_to_webhead__http"
-    description = "Allow HTTP(S) from ELBs to webhead."
-    ingress {
-        from_port = 80
-        to_port = 80
-        protocol = "tcp"
-        security_groups = [
-            "${aws_security_group.internet_to_elb__http.id}"
-        ]
-    }
-    ingress {
-        from_port = 443
-        to_port = 443
-        protocol = "tcp"
-        security_groups = [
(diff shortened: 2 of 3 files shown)


phrawzty:
*how do we deal with DNS? We want this to be http://symbolapi.mozilla.org*
We'll have to bump this question up the ladder; I'm thinking webops, JP, and releng might be able to provide is with the info.

*should there be an output variable for this (the public IP?)*
Why not?

*would it be valuable to use an ELB even for a single node like this? Would that e.g. enable us to automatically bring up a new VM if this one failed?*
"Automatically" is a strong word :smirk:  - there is additional additional work to do before we get to that point. Furthermore, there is a cost associated with ELB, so that's a factor as well. Right now my guess is that we *don't* need an ELB for a single instance, but that's not written in stone...

phrawzty:
We *definitely* need to explain how we're using branches in the README.

rhelmer:
To summarize discussion from IRC:

1) we could use an ELB, not shut it down, and (optionally) auto-scaling groups in case the node dies

The "not shutting it down" part makes me nervous. Alternatively:

2) use an Elastic IP pointing to the instance (you can't point these at ELBs)

This means we can destroy/apply at will, but if the node dies nothing will automatically pick it back up.
I think that 1 is what we're going to end up needing to do with other services anyway, so we might as well figure it out for this relatively low-stakes service.

terraform/README.md
```diff
@@ -27,3 +27,9 @@ keys in `terraform.tfvars` and specifying the rest at runtime:
 terraform plan -var 'environment=foo' -var 'del_on_term=true' -out=foo.plan
 terraform apply foo.plan
 ```
+
+## Branches
+
+The ```master``` branch contains production-ready configs, changes that are
```
phrawzty: `ready configs; changes`
grammar_hammer.gif

phrawzty:
Are we going to implement an ELB in this PR or ?

rhelmer:
@phrawzty yes I'll go ahead and implement an ELB in this PR, and get the public IP as an output variable. We can use that for our CNAME, and just deal with DNS if we ever need to destroy the ELBs. :fire:

rhelmer:
@phrawzty hmm you know what, I am going to go make these changes on the stage branch, then cherry-pick then over here, so the two don't get out of sync. I'll open a separate PR for that.

rhelmer:
OK per https://github.com/mozilla/socorro-infra/pull/18#issuecomment-76204545 we're not going to havea ```stage``` integration branch, ```master``` will be shippable stuff from now on.

terraform/README.md
```diff
@@ -27,3 +27,9 @@ keys in `terraform.tfvars` and specifying the rest at runtime:
 terraform plan -var 'environment=foo' -var 'del_on_term=true' -out=foo.plan
 terraform apply foo.plan
 ```
+
+## Branches
+
+The ```master``` branch contains production-ready configs; changes that are
```
phrawzty: This will need to be re-worded.

terraform/README.md
```diff
@@ -27,3 +27,8 @@ keys in `terraform.tfvars` and specifying the rest at runtime:
 terraform plan -var 'environment=foo' -var 'del_on_term=true' -out=foo.plan
 terraform apply foo.plan
 ```
+
+## Making changes live in staging and production
+
+Anything landed on master will go to staging and be considered eligible for
+pushing to production. Tags will be pushed to production.
```
rhelmer: @phrawzty do you think it's worth mentioning this here? Purposely not mentioning that it's automated or manual, since presumably this will be manual the first few times until we automate it.

phrawzty: :+1: for now

phrawzty:
`r+` :cake:

rhelmer:
:bomb:s away

phrawzty:
it's happening

PR #16, opened 2015-02-25
Bug1127532 symbolapi stage

rhelmer:
r? @phrawzty - let's get the settings for PR #15 right on the stage branch first, and then I'll cherry-pick the parts that are ready over to master.

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -30,6 +30,6 @@ terraform apply foo.plan
 
 ## Branches
 
-The ```master``` branch contains production-ready configs, changes that are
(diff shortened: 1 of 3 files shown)


phrawzty:
Edits look good; however, before merging I want to review our management strategy.  Some questions:

* Would it make more sense to propose this PR against the `stage` branch directly, instead of `master`?
* Since we're using branches to manage environments, would it make sense to have a `prod` branch also?
* Would it then make more sense for `master` to effectively be a `dev` branch, with merges to `stage` and/or `prod` affecting those environments explicitly?

PR #18, opened 2015-02-26
Bug1127532 symbolapi stage

rhelmer:
This is what I mean to do in PR #16 - open against the stage branch.

@phrawzty re: your comments in that one, having ```master``` branch be for dev and a separate ```prod``` branch is not a bad idea.... I'm not sure? :) Let's think/talk about it more, there are some advantages to having ```master``` be production too.

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -30,6 +30,6 @@ terraform apply foo.plan
 
 ## Branches
 
-The ```master``` branch contains production-ready configs, changes that are
+The ```master``` branch contains production-ready configs; changes that are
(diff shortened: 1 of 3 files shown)


rhelmer:
OK discussed this in IRC - let's not have ```stage``` integration branch at all. We'll have a temporary ```demo``` branch to hold the work done so far, but work will just land on ```master``` via PRs as it's ready.

Once we have automation going, pushes to ```master``` will auto-update a staging environment and tags will push to production, but we're not quite there yet.

phrawzty:
:+1:

PR #19, opened 2015-02-26
bug 1127532 - get symbolapi working

rhelmer:
A few things I missed on the initial landing, after more testing.


phrawzty:
`r+` :cactus:

PR #22, opened 2015-03-03
update AMI ID

phrawzty:
Now with the Socorro public repo by default!

@rhelmer `r?`


rhelmer:
r+++ would review again

PR #24, opened 2015-03-04
bug 1127532 - add ASG and launch config for symbolapi.mozilla.org

rhelmer:
r? @phrawzty


phrawzty:
:a: :heavy_plus_sign: WOULD REVIEW AGAIN

PR #27, opened 2015-03-04
move puppet

phrawzty:
We're going to start using Puppet (masterless) as the provisioning mechanism at the ASG (via `user-data`) phase, so it makes sense to move the directory out of `packer/` and make it a primary resource.

@rhelmer `r?`


rhelmer:
r+ lgtm and wfm

PR #28, opened 2015-03-05
use Puppet for role provision

phrawzty:
### Details

This is a reasonably large set of diffs, so let's walk through the important bits. :dancers:

**`puppet/manifests/default.pp`**
* Previously this was only used during the Packer phase; I've added another block for the "role" phase (ASG provisioning).

**`puppet/modules/socorro/manifests/init.pp`**
* Previously, Git was only used by the Buildbox; however, it is now used to pull down the latest master of `socorro-infra`, so that the role can be provisioned.

**`puppet/modules/socorro/manifests/packer/*.pp`**
* With the exception of `Package['git']` these haven't changed - they've just been moved.

**`puppet/modules/socorro/manifests/role/consul.pp`**
* This sets up the [Consul](https://consul.io/) nodes with a minimum viable configuration.
* The service starts but will not form a cluster - there is a manual bootstrap that has to be performed. It's not so bad.

**`puppet/modules/socorro/manifests/role/symbolapi.pp`**
* This replaces `symbolapi_cloud-init.sh` (yay).

**`terraform/main.tf`**
* I've re-added the `private_to_private__any` rule. Consul uses five different ports and both TCP and UDP, so, this was easier. We can (and should) review our internal ACLs as a future optimisation.
* Line `107` (and similar on `121`) has this little gem of a hack:

    ```ini
    user_data = "${file(\"socorro_role.sh\")} ${var.infra_repo} symbolapi"
    ```
    This causes `socorro_role.sh` to be imported into `user-data` as expected, but then *concatenates* `${var.infra_repo} symbolapi` to the end of that script. Keep this in mind when we examine the script itself.
* It would appear that the default health check for ASG resources is `ELB`; I therefore had to explicitly set `health_check_type = "EC2"` in the `asg_for_consul` resource since it doesn't use an ELB.  We'll have to keep this in mind because it's easy to miss.

**`terraform/socorro_role.sh`**
* The `socorro_role` function is meant to clone `socorro-infra` and then apply the appropriate role - but where is it getting `$1` and `$2` from?  The function call on the final line contains a slash such that further input is expected - the "missing" next line is supplied by Terraform in the `user_data` definition.
* I can't tell whether this is super clever, super dirty, or both. :stuck_out_tongue_closed_eyes: 

**`terraform/variables.tf`**
* The actual location of the infra repo can be over-written in case one wanted to try out a different repo (for dev purposes, for example).

### Known bug!

The previous base AMI does *not* have `git` installed, which means that the provisioning step will currently fail. Solution:
* Merge this branch.
* Roll a new AMI from this branch.
* Update `variables.tf` with the new ID.
* Prepare PR and merge.

### Future improvements

* Add "environmental awareness" to `puppet/modules/socorro/manifests/init.pp`: have this download a *tagged release* of `socorro-infra` if we're Prod and pull master if we're stage.

### r?

@rhelmer :two_hearts:

### DIFF
--- a/terraform/socorro_role.sh
+++ b/terraform/socorro_role.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+DIR="/tmp/${RANDOM}-${RANDOM}"
+
+function socorro_role {
+    git clone $1 $DIR
+    /usr/bin/env FACTER_socorro_role=$2 \
+        puppet apply \
+        --modulepath=${DIR}/puppet/modules \
+        ${DIR}/puppet/manifests/default.pp
+}
+
+# Required variables will be inserted by Terraform.
+socorro_role \
--- a/terraform/socorro_role.sh
+++ b/terraform/socorro_role.sh
@@ -3,11 +3,17 @@
 DIR="/tmp/${RANDOM}-${RANDOM}"
 
 function socorro_role {
-    git clone $1 $DIR
+    mkdir $DIR
+    pushd $DIR
+    curl -O $1
+    # Yoink the filename from the end of the URL
+    ARCHIVE=`echo $1|awk -F'/' '{print $NF}'`
+    tar -xvzf $ARCHIVE
     /usr/bin/env FACTER_socorro_role=$2 \
         puppet apply \
         --modulepath=${DIR}/puppet/modules \
         ${DIR}/puppet/manifests/default.pp
+    popd
 }
 
 # Required variables will be inserted by Terraform.
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -22,6 +22,6 @@ variable "base_ami" {
 variable "alt_ssh_port" {
     default = 22123
 }
-variable "del_on_term" {
-    default = "false"
+variable "infra_repo" {
+    default = "https://github.com/mozilla/socorro-infra.git"
 }
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -22,6 +22,6 @@ variable "base_ami" {
 variable "alt_ssh_port" {
     default = 22123
 }
-variable "infra_repo" {
-    default = "https://github.com/mozilla/socorro-infra.git"
+variable "puppet_archive" {
+    default = "https://s3-us-west-2.amazonaws.com/org.mozilla.crash-stats.packages-public/prov_cache/socorro-infra__puppet.tar.gz"
 }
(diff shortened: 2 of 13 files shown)


terraform/socorro_role.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+DIR="/tmp/${RANDOM}-${RANDOM}"
+
+function socorro_role {
+    git clone $1 $DIR
```
rhelmer: One thing that gives me pause here is that we're making spinning up instances (even unattended, by ASGs) dependent on github.com being up and working. It's one thing if we can't deploy a new version of the software because GitHub is down, it's quite another to not be able to auto-scale or otherwise spin up new instances.

rhelmer: I think it'd be safer to have the puppet manifests in the AMI - means that adjusting roles requires packing a new AMI and switching to it, but at least we wouldn't be blocking spinning up new instances on any remote services.

phrawzty: Your criticism of the GitHub reliance is a fair one, and it's absolutely something we should address a little further down the line - right now, building new AMIs every time the socorro-infra repo changes is onerous (since it's manual), but when we get into a position where this is happening automatically (from CI, for example), it will make more sense.

Of course, if we're building new base AMIs all the time, we also end up running into a situation where we have subtly different nodes all over the place (recall that part of the provisioning step is to run a `yum upgrade`) - this is a bad smell, and is exactly why we wanted to use longer-lived base AMIs in the first place.

Perhaps there's a middle path?  How about keeping a copy of socorro-infra somewhere in Amazon, such as S3, or even on the Admin box?  I'm not sure exactly what form this would take, but it would satisfy both issues: not reliant on GH, and doesn't require AMIs to be rebuilt all the time.

rhelmer: @phrawzty this all makes sense - I think S3 is probably more reliable for this than GH.

terraform/variables.tf
```diff
@@ -22,6 +22,6 @@ variable "base_ami" {
 variable "alt_ssh_port" {
     default = 22123
 }
-variable "del_on_term" {
-    default = "false"
+variable "puppet_archive" {
```
phrawzty: This requires that we maintain up-to-date tarball of the Puppet dir from `socorro-infra` - luckily, this is easy to automate. :smile:

rhelmer:
r+, I'm sure we'll revisit in the future but seems reasonable for now.

PR #29, opened 2015-03-10
fix bug 1118297 - terraform config for Socorro

rhelmer:
As we discussed, ```main.tf``` is getting ungainly with so many resources, so we're probably going to want to split this up (multiple ```.tf``` files and/or use terraform modules for some of the common pattern here, like the ELB+ASG+LC combo we do for almost everything), but I think this is a good enough start.

The rule of thumb I have used here is that plain instances are only used if the instances aren't ephemeral, so ES (which uses instance storage) and PG (which uses EBS), these two are sensitive to data loss (it'll take quite a while to recover.) Therefore these two services need to be monitored and recovered manually, every other service should be able to auto-recover.

Besides testing this more, I'd like to figure out why I get ```Error retrieving AutoScaling groups: Throttling: Rate exceeded``` when I ```terraform destroy``` before we merge this - presumably terraform (v0.3.7) is querying the AWS API too quickly during ```destroy```.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -69,6 +69,7 @@ resource "aws_security_group" "internet_to_elb__http" {
     }
 }
 
+# symbolapi
 resource "aws_security_group" "elb_to_symbolapi__http" {
     name = "${var.environment}__elb_to_symbolapi__http"
     description = "Allow HTTP(S) from ELBs to symbolapi."
@@ -116,18 +117,6 @@ resource "aws_launch_configuration" "lc_for_symbolapi_asg" {
     ]
 }
 
-resource "aws_launch_configuration" "lc_for_consul_asg" {
-    name = "${var.environment}__lc_for_consul_asg"
-    user_data = "${file(\"socorro_role.sh\")} ${var.puppet_archive} consul"
-    image_id = "${lookup(var.base_ami, var.region)}"
-    instance_type = "t2.micro"
-    key_name = "${lookup(var.ssh_key_name, var.region)}"
-    security_groups = [
-        "${aws_security_group.internet_to_any__ssh.name}",
-        "${aws_security_group.private_to_private__any.name}"
-    ]
-}
-
 resource "aws_autoscaling_group" "asg_for_symbolapi" {
     name = "${var.environment}__asg_for_symbolapi"
     availability_zones = [
@@ -146,6 +135,18 @@ resource "aws_autoscaling_group" "asg_for_symbolapi" {
     ]
 }
 
+resource "aws_launch_configuration" "lc_for_consul_asg" {
+    name = "${var.environment}__lc_for_consul_asg"
+    user_data = "${file(\"socorro_role.sh\")} ${var.puppet_archive} consul"
+    image_id = "${lookup(var.base_ami, var.region)}"
+    instance_type = "t2.micro"
+    key_name = "${lookup(var.ssh_key_name, var.region)}"
+    security_groups = [
+        "${aws_security_group.internet_to_any__ssh.name}",
+        "${aws_security_group.private_to_private__any.name}"
+    ]
+}
+
 resource "aws_autoscaling_group" "asg_for_consul" {
     name = "${var.environment}__asg_for_consul"
     availability_zones = [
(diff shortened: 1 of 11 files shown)


terraform/main.tf
```diff
@@ -161,3 +162,387 @@ resource "aws_autoscaling_group" "asg_for_consul" {
...
+    instance_type = "t2.micro"
+    key_name = "${lookup(var.ssh_key_name, var.region)}"
+    count = 1
+    security_groups = [
+        "${aws_security_group.internet_to_any__ssh.name}",
+        "${aws_security_group.private_to_private__any.name}"
+    ]
+    block_device {
+        device_name = "/dev/sda1"
+        delete_on_termination = "${var.del_on_term}"
```
phrawzty: I don't think that this is a good place for a global boolean - just set this to `false` imho.

terraform/variables.tf
```diff
@@ -25,3 +25,6 @@ variable "alt_ssh_port" {
 variable "puppet_archive" {
     default = "https://s3-us-west-2.amazonaws.com/org.mozilla.crash-stats.packages-public/prov_cache/socorro-infra__puppet.tar.gz"
 }
+variable "del_on_term" {
```
phrawzty: I had this in the demo branch only as a convenient way to clean up test instances - I don't think it's appropriate for a "real" config.

phrawzty:
> As we discussed, main.tf is getting ungainly with so many resources, so we're probably going to want to split this up (multiple .tf files and/or use terraform modules for some of the common pattern here, like the ELB+ASG+LC combo we do for almost everything), but I think this is a good enough start.

The advantage to *not* using modules is that it keeps things basic, which has value; conversely, since we're repeating same pattern over and over, it's hard to justify *not* using a module in the first place.  In any case, I agree with you on all counts: this is "good enough" for now, but we'll definitely want to invest in either splitting it up or using modules (possibly both, since they aren't mutually exclusive). 

> The rule of thumb I have used here is that plain instances are only used if the instances aren't ephemeral, so ES (which uses instance storage) and PG (which uses EBS), these two are sensitive to data loss (it'll take quite a while to recover.) Therefore these two services need to be monitored and recovered manually, every other service should be able to auto-recover.

This is also in line with what we discussed.  That said, with regards to ES, since it just uses instance storage, and since the service heals itself from node losses (assuming we're using replication), there's a strong argument to be made for letting ASG handle it as well.  There are other issues at play regarding ES as well: for example, whether we're going to use small number of beefy nodes or a larger number of small nodes, and whether we're going to employ a role-based model or not.  These questions play *directly* into our infrastructure strategy and thus have ramifications in Terraform.

Concerning PG, yes, it's a bit of a special case and should be treated as such - my hope is that we're not going to have to support it directly in the long-term (read: 3rd-party).

> Error retrieving AutoScaling groups: Throttling: Rate exceeded

Yeah that's a weird one.  Let's repro, gather some data, and then hit up the mailing list and/or submit an issue upstream.

phrawzty:
The Puppet stubs (which desperately need linting :smirk:) are a good starting point.  We're also now starting to straddle the line between a Puppet module for internal use and something that is clearly larger than that - for example, we're provisioning a raft of services for which there are already Puppet modules out there.  We're going to have to make some policy decisions before we end up digging ourselves into a hole.

rhelmer:
OK regarding the AWS throttling issue, looks like it's already filed: https://github.com/hashicorp/terraform/issues/1051

As mentioned in the issue it's not the right long-term solution, but I am going to see if setting parallelism down helps resolve the problem for now.

rhelmer:
@phrawzty ok removed the ```del_on_term``` stuff (although I think it is really excellent for development!)

> The Puppet stubs (which desperately need linting :smirk:) are a good starting point. We're also now starting to straddle the line between a Puppet module for internal use and something that is clearly larger than that - for example, we're provisioning a raft of services for which there are already Puppet modules out there. We're going to have to make some policy decisions before we end up digging ourselves into a hole.

As of 4f50816 travis yells if you fail puppet-lint ;)

Hm most of these role ```.pp``` files are very tiny, things like making sure a package or two is installed and a service or two is started. Can you elaborate on which module(s) would help out here?

phrawzty:
> @phrawzty ok removed the del_on_term stuff (although I think it is really excellent for development!)

As we discussed on IRC, it's great for development, and I've no problem with keeping it in there, so long as we document the crap out of it (because it's potentially dangerous).  At this point it's a nit though - we can add it back in later if we want.

> Hm most of these role .pp files are very tiny, things like making sure a package or two is installed and a service or two is started. Can you elaborate on which module(s) would help out here? 

The *package / config / service* model is simple and well known in Puppet, and I'm not advocating that we use modules right away, but it's a fine line to walk between "simple enough to do ourselves" and "oh god we've created a monster". :smiling_imp: That said, in the interest of moving things along, I'm willing to shelve this topic for today.

That said, some of them will have to grow right away, since we need to ship configuration files for things like Elasticsearch, RabbitMQ, Apache, and so forth.  I propose that we merge this PR and then continue to flesh out the individual roles afterwards.

rhelmer:
@phrawzty OK reverted that "remove_del_on_term" removal, will rebase and land this in that case.

phrawzty:
`yarrrrr+`

PR #30, opened 2015-03-10
run travis to test terraform and puppet configs

rhelmer:


phrawzty:
:+1: to adding Terraform and Puppet tests

rhelmer:
r? @phrawzty - OK this seems to work now, there isn't a ```terraform lint``` yet (although there's an open issue for it), ```terraform graph``` doesn't require auth and catches syntax errors as well as problems in the graph (cycles etc) so seems like a good thing for now.

At least we'll be notified in PRs if we try to check in any syntax errors!

.travis.yml
```diff
@@ -0,0 +1,11 @@
+sudo: no
+language: python
+before_install:
+  - gem install puppet puppet-lint
+  - cd terraform; wget 'https://dl.bintray.com/mitchellh/terraform/terraform_0.3.7_linux_amd64.zip'; unzip -u terraform_0.3.7_linux_amd64.zip; cd ../
+
+install: false
+
+script:
+  - puppet-lint --with-filename puppet/
```
phrawzty: May wish to add parser validation too (i.e. `puppet parser validate <file>`).  I'm not sure what the best way to do this would be.  Could iterate from `find`, for example:
```
find puppet/ -type f -name '*.pp' -exec puppet parser validate {} \;
```
This is straightforward but it requires the binary to be loaded each time. :turtle:  Something like this is less simple, but *much* faster:
```
find puppet/ -name '*.pp' | xargs echo > out && puppet parser validate `cat out`
```
In the future we may wish to do *actual* Puppet tests, but that's really out of scope for now.

.travis.yml
```diff
@@ -0,0 +1,11 @@
+sudo: no
+language: python
+before_install:
+  - gem install puppet puppet-lint
+  - cd terraform; wget 'https://dl.bintray.com/mitchellh/terraform/terraform_0.3.7_linux_amd64.zip'; unzip -u terraform_0.3.7_linux_amd64.zip; cd ../
```
phrawzty: We'll eventually need a way to ensure that we're testing against the same version that we're deploying with - not critical now, but let's put it on the roadmap.

.travis.yml
```diff
@@ -0,0 +1,12 @@
...
+
+install: false
+
+script:
+  - puppet-lint --with-filename puppet/
+  - puppet parser validate `find puppet/ -name '*.pp'`
+  - cd terraform; ./terraform plan -var="environment=blah" -var="secret_key=FAKE" -var="access_key=FAKE"
```
phrawzty: Prefer the `pushd` `popd` pattern to `cd` since it's more explicit, and it helps to avoid problems where we `cd` in one step and forget to back out in the next.

.travis.yml
```diff
@@ -0,0 +1,12 @@
+sudo: no
+language: python
+before_install:
+  - gem install puppet puppet-lint
+  - pushd terraform; wget 'https://dl.bintray.com/mitchellh/terraform/terraform_0.3.7_linux_amd64.zip'; unzip -u terraform_0.3.7_linux_amd64.zip; popd
```
phrawzty: Does `terraform/` exist beforehand somehow?

.travis.yml
phrawzty: Also, I'm pretty sure that `terraform plan` will work if we just supply fake vars. :smile:

.travis.yml
rhelmer: @phrawzty yep it's the ```terraform/``` dirs in the root of the ```socorro-infra``` repo

phrawzty: Oh, right! In that case I don't know if we should be unzipping here, but this is a nit and I'll not block on it. :smile:

.travis.yml
phrawzty: I would suggest adding `--no-80chars-check --no-autoloader_layout-check` as args to `puppet-lint` since that's what we use elsewhere.

phrawzty:
### `r+++++` would review again

PR #31, opened 2015-03-11
fix bug 1142067; add buildbox role

phrawzty:
DISCLAIMER: I haven't tested this yet.  Just wanted to get it up so we could take a look.  cc @rhelmer

### DIFF
--- a/puppet/modules/socorro/manifests/packer/base.pp
+++ b/puppet/modules/socorro/manifests/packer/base.pp
@@ -12,10 +12,7 @@ class socorro::packer::base {
     'postgresql-9.3':
       ensure  => stopped,
       enable  => false,
-      require => [
-        Package['postgresql93-server'],
-        File['pg_hba.conf'],
-      ];
+      require => Package['postgresql93-server'];
 
     'elasticsearch':
       ensure  => stopped,
@@ -81,15 +78,6 @@ class socorro::packer::base {
   file {
     '/etc/socorro':
       ensure => directory;
-
-    'pg_hba.conf':
-      ensure  => file,
-      path    => '/var/lib/pgsql/9.3/data/pg_hba.conf',
-      source  => 'puppet:///modules/socorro/var_lib_pgsql_9.3_data/pg_hba.conf',
-      owner   => 'postgres',
-      group   => 'postgres',
-      require => Package['postgresql93-server'],
-      notify  => Service['postgresql-9.3'];
 
     'pgsql.sh':
       ensure => file,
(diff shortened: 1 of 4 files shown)


puppet/modules/socorro/manifests/role/buildbox.pp
```diff
@@ -0,0 +1,72 @@
...
+    'rabbitmq-server':
+      ensure  => running,
+      enable  => true;
+
+    'postgresql-9.3':
+      ensure  => running,
+      enable  => true,
+      require => Exec['postgres-initdb'];
+
+    'elasticsearch':
```
rhelmer: we discussed in IRC, but probably worth calling out here - it looks like our ```socorro/unittest/external/elasticsearch``` tests fail when run with ES >= 1.0 :(

This is frustrating because we need to continue running on 0.9.x in PHX and want to upgrade to 1.x in AWS, so ideally we could run tests for both but I think until we get the test situation ironed out we should just run tests with 0.9.x for now.

phrawzty: There is a better way: be more specific about which tests we run based on which version of ES is in play.  We already have this pattern in place, but it's not complete - to wit, I've filed [bug 1142475](https://bugzilla.mozilla.org/show_bug.cgi?id=1142475), which @AdrianGaudebert and I will work on.

puppet/modules/socorro/manifests/packer/base.pp
```diff
@@ -82,15 +79,6 @@
     '/etc/socorro':
       ensure => directory;
 
-    'pg_hba.conf':
```
phrawzty: Removing this has repercussions: we'll need to manage it in any eventual role that needs it.  The good news is that, in principle, this would just be a Postgres server (which we might not end up running), and Buildbox.

phrawzty:
@rhelmer `r?`

puppet/modules/socorro/manifests/role/buildbox.pp
```diff
@@ -0,0 +1,80 @@
...
+  exec {
+    'postgres-initdb':
+      path    => '/usr/bin:/usr/sbin',
+      command => '/usr/pgsql-9.3/bin/postgresql93-setup initdb'
+  }
+
+  exec {
+    'postgres-trust-local':
+      path    => '/usr/bin',
+      command => 'sed -i "s:ident:trust:" /var/lib/pgsql/9.3/data/pg_hba.conf',
```
rhelmer: ಠ_ಠ

rhelmer: Do you think it'd be better to just carry around a copy of this file, or use templates?

phrawzty: Either of those would be better, frankly.  This was purely a question of expediency - we're running out of time in the Q. :sweat:

rhelmer: OK I think this is fine as it's just for tests on the buildbox, if it breaks we can fix.

rhelmer:
r+

PR #34, opened 2015-03-17
add golang and mercurial to build deadci

phrawzty:
Necessary in order to build DeadCI from source (which we have to do until such a time as my [PR](https://github.com/phayes/deadci/pull/2) is accepted).

@rhelmer `r?`


rhelmer:
r+

phrawzty:
Given that we're no longer using DeadCI, I suppose we could (should?) remove golang and mercurial...

rhelmer:
@phrawzty golang we'll need to build terraform/consul RPMs right?

phrawzty:
> @phrawzty golang we'll need to build terraform/consul RPMs right?

Nope - the RPMs are built from the release tarballs.

PR #35, opened 2015-03-18
fix bug 1144130; add hiera-s3

phrawzty:
This ensures that [hiera-s3](https://github.com/phrawzty/hiera-s3) is available from base.

I also took the liberty of removing two stale manifests from the repo.

@rhelmer `r?`


rhelmer:
r+

PR #36, opened 2015-03-19
add buildbox elements

phrawzty:
This will "work" in that the AWS resources will come up, but the DeadCI config will be populated with nil values since we don't actually have a Hiera backend yet. :vhs: 

@rhelmer `r?`

### DIFF
--- a/puppet/modules/socorro/templates/var_lib_deadci/deadci.ini.erb
+++ b/puppet/modules/socorro/templates/var_lib_deadci/deadci.ini.erb
@@ -1,0 +1,8 @@
+command = <%= @deadci_command %>
+port = <%= @deadci_port %>
+#host = <%= scope['::ec2_public_hostname'] %>
+httpsclone = true
+[github]
+enabled = true
+token = <%= @deadci_token %>
+secret = <%= @deadci_hmac %>
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -55,7 +55,7 @@ resource "aws_security_group" "internet_to_any__ssh" {
 
 resource "aws_security_group" "internet_to_elb__http" {
     name = "${var.environment}__internet_to_elb__http"
-    description = "Allow incoming traffic from Internet to HTTP(S) on ELBs."
+    description = "Allow incoming traffic from Internet to HTTP on ELBs."
     ingress {
         from_port = 80
         to_port = 80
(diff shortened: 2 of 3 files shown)


terraform/main.tf
```diff
@@ -55,7 +55,7 @@ resource "aws_security_group" "internet_to_any__ssh" {
 
 resource "aws_security_group" "internet_to_elb__http" {
     name = "${var.environment}__internet_to_elb__http"
-    description = "Allow incoming traffic from Internet to HTTP(S) on ELBs."
+    description = "Allow incoming traffic from Internet to HTTP on ELBs."
```
rhelmer: I think we do want to reintroduce HTTPS at some point! For now fixing comment is ok

rhelmer:
r+

puppet/modules/socorro/templates/var_lib_deadci/deadci.ini.erb
```diff
@@ -0,0 +1,8 @@
+command = <%= @deadci_command %>
+port = <%= @deadci_port %>
+#host = <%= scope['::ec2_public_hostname'] %>
```
phrawzty: Leaving this commented since I'm not sure what exactly to put here (or even if we need to worry about it, tbh).

puppet/modules/socorro/templates/var_lib_deadci/deadci.ini.erb
```diff
@@ -0,0 +1,8 @@
...
+httpsclone = true
```
phrawzty: Conceivably we could run more than one DeadCI instance (different projects? idk.) so this could be made a variable as well.  THE FUTURE.

terraform/main.tf
```diff
@@ -540,3 +540,69 @@ resource "aws_elb" "elb_for_elasticsearch" {
         "${aws_instance.elasticsearch.*.id}"
     ]
 }
+
+# Buildbox
+resource "aws_security_group" "elb_to_buildbox__http" {
+    name = "${var.environment}__elb_to_buildbox__http"
+    description = "Allow HTTP from ELBs to buildbox."
+    ingress {
+        from_port = 80
```
phrawzty: Man, I sure hope that DeadCI doesn't have any obvious remotely exploitable flaws! :skull:

rhelmer: This is for the webook for github to call?

phrawzty: Correct.

PR #39, opened 2015-03-22
set up puppet so launch configs work, including using singular name for ...

rhelmer:
...roles consistently

r? @phrawzty - I've only tested collector so far, which seems to WFM with these changes.

### DIFF
--- a/puppet/modules/socorro/manifests/role/admin.pp
+++ b/puppet/modules/socorro/manifests/role/admin.pp
@@ -6,4 +6,12 @@ class socorro::role::admin {
       ensure=> latest
   }
 
+  file {
+    '/etc/cron.d/socorro':
+      mode   => '0600',
+      owner  => root,
+      group  => root,
+      source => 'puppet:///modules/socorro/etc_cron.d/socorro'
+  }
+
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ before_install:
 install: false
 
 script:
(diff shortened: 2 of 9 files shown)


puppet/modules/socorro/manifests/role/admin.pp
```diff
@@ -6,4 +6,12 @@
       ensure=> latest
   }
 
+  file {
+    '/etc/cron.d/socorro':
+      mode => 600,
+      owner => root,
+      group => root,
+      source => 'puppet:///modules/socorro/files/etc_cron.d/socorro'
```
rhelmer: Will this work ok with masterless puppet? Not sure how to specify the path here if not (obv. I haven't tested this role either :) )

rhelmer:
@phrawzty OK I read up on it a bit and fixed up the admin role. I went ahead and tested all socorro-specific roles (everything but ES, PG, Consul) and this seems to work fine for all of them, with the latest socorro RPM in the repo (which is build from my ```centos7-fpm-update``` branch) \o/

phrawzty:
Looks good.  There are just some style nits to fix, though: lines 140-144 of the [Travis run](https://travis-ci.org/mozilla/socorro-infra/builds/55336749) are `puppet-lint` warnings.  Please to be fixing those before merge. :page_facing_up: Otherwise `r+`.

phrawzty:
Also: suggest adding `--fail-on-warnings` as an arg to `puppet-lint` in `.travis.yml`.

PR #40, opened 2015-03-24
bug 1146537; adjust cfg dir (upstream), split consul client & server

phrawzty:
This splits out the configs for Consul clients and servers (previously we only dealt with servers) by putting the client portions in Base and the server portions (only) in the appropriate role.

As a side-effect, this also fixes [bug 1146537](https://bugzilla.mozilla.org/show_bug.cgi?id=1146537), which addresses a change in the config dir path.

This will require the Base and Buildbox AMIs to be rebuilt.

@rhelmer `r?`


rhelmer:
r+

PR #41, opened 2015-03-24
fix bug 1146449; pass the secret bucket name to Hiera on provision

phrawzty:
I am extending the pattern of passing arbitrary information (in this case, the secret bucket name) via `user_data`. This value is then picked up (positionally) by `socorro_role.sh` and applied to `hiera.yaml`, thus allowing us to configure `hiera-s3` without exposing the name of the secret bucket.  Once again, I'm not sure if this is clever or janky - possibly both - but it should work. :smiling_imp: 

This also touches on [bug 1146447](https://bugzilla.mozilla.org/show_bug.cgi?id=1146447).

@rhelmer `r?`

### DIFF
--- a/puppet/modules/socorro/files/etc_puppet/hiera.yaml
+++ b/puppet/modules/socorro/files/etc_puppet/hiera.yaml
@@ -1,0 +1,14 @@
+---
+:backends:
+    - consul
+    - S3
+
+:consul:
+    :host: 127.0.0.1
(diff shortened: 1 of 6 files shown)


phrawzty:
This will require both a rebuild of the Base and Buildbox AMIs, as well as an update of the Puppet provision archive.

rhelmer:
r+

PR #44, opened 2015-03-26
fixes for consul in puppet

phrawzty:
@rhelmer `r?`

### DIFF
--- a/puppet/modules/socorro/files/etc_consul/common.json
+++ b/puppet/modules/socorro/files/etc_consul/common.json
@@ -1,4 +1,4 @@
 {
     "data_dir": "/var/lib/consul",
-    "log_level": "INFO",
+    "log_level": "INFO"
 }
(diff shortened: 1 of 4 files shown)


puppet/modules/socorro/files/etc_consul/ui.json
```diff
@@ -1,3 +0,0 @@
-{
```
phrawzty: The new `consul-ui` package ships with this file so we don't need to supply it here.

rhelmer:
r+

puppet/modules/socorro/files/etc_consul/common.json
```diff
@@ -1,4 +1,4 @@
 {
     "data_dir": "/var/lib/consul",
-    "log_level": "INFO",
```
phrawzty: This is the reason humans aren't supposed to edit JSON by hand. :black_medium_square:

terraform/variables.tf
```diff
@@ -17,7 +17,7 @@ variable "ssh_key_name" {
 }
 variable "base_ami" {
     default = {
-        us-west-2 = "ami-2fbc931f"
+        us-west-2 = "ami-51604e61"
```
phrawzty: I have already built a new base AMI from this PR. :smile:

rhelmer: AMIception

PR #45, opened 2015-03-26
no need for standalone middleware nodes, run socorro-middleware service ...

rhelmer:
...on webapp nodes instead


phrawzty:
`r+`

PR #46, opened 2015-03-26
tighten up buildbox config and fix modulepath

phrawzty:
@rhelmer `r?`

### DIFF
--- a/puppet/modules/socorro/manifests/role/buildbox.pp
+++ b/puppet/modules/socorro/manifests/role/buildbox.pp
@@ -39,7 +39,7 @@ class socorro::role::buildbox {
   file {
     'deadci.ini':
       ensure  => file,
-      path    => '/var/lib/deadci.ini',
+      path    => '/var/lib/deadci/deadci.ini',
       content => template('socorro/var_lib_deadci/deadci.ini.erb'),
       owner   => 'deadci',
       group   => 'deadci',
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -126,7 +126,6 @@ resource "aws_launch_configuration" "lc_for_symbolapi_asg" {
     instance_type = "c4.xlarge"
     key_name = "${lookup(var.ssh_key_name, var.region)}"
     security_groups = [
-        "${aws_security_group.internet_to_elb__http.name}",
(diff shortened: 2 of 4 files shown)


puppet/modules/socorro/manifests/role/buildbox.pp
```diff
@@ -39,7 +39,7 @@
   file {
     'deadci.ini':
       ensure  => file,
-      path    => '/var/lib/deadci.ini',
+      path    => '/var/lib/deadci/deadci.ini',
```
phrawzty: :blush:

terraform/socorro_role.sh
```diff
@@ -18,7 +18,7 @@ function socorro_role {
     # Provision the role.
     /usr/bin/env FACTER_socorro_role=$2 \
         puppet apply \
-        --modulepath=${DIR}/puppet/modules \
+        --modulepath=${DIR}/puppet/modules:/etc/puppet/modules \
```
phrawzty: We need both *our* modules as well as those in the system.

terraform/variables.tf
```diff
@@ -20,6 +20,11 @@ variable "base_ami" {
         us-west-2 = "ami-51604e61"
     }
 }
+variable "buildbox_ami" {
```
phrawzty: Time to start tracking this, too.

terraform/main.tf
```diff
@@ -471,28 +481,28 @@ resource "aws_elb" "elb_for_buildbox" {
...
     instance_type = "t2.micro"
     key_name = "${lookup(var.ssh_key_name, var.region)}"
     security_groups = [
-        "${aws_security_group.internet_to_elb__http.name}",
-        "${aws_security_group.elb_to_buildbox__http.name}",
+        "${aws_security_group.elb_to_buildbox__deadci.name}",
         "${aws_security_group.internet_to_any__ssh.name}",
         "${aws_security_group.private_to_private__any.name}"
     ]
+    iam_instance_profile = "buildbox"
```
phrawzty: This is the magic right here. We'll need to apply `iam_instance_profile` throughout.  By default we'll use the `default` role. :wine_glass:

rhelmer:
r+

PR #47, opened 2015-03-27
bug 1148302 - run crashmover on collectors

rhelmer:
? @phrawzty - this only partially addresses bug 1148302, it's what the configs in ```socorro-config``` are expecting though (collection to disk, crashmover service to move those into S3+Rabbit)

### DIFF
--- a/puppet/modules/socorro/manifests/role/collector.pp
+++ b/puppet/modules/socorro/manifests/role/collector.pp
@@ -11,6 +11,11 @@ class socorro::role::collector {
       ensure  => running,
       enable  => true,
       require => Package['socorro'];
+
+    'socorro-crashmover':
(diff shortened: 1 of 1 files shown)


phrawzty:
```
09:59:15 < phrawzty> rhelmer: https://github.com/mozilla/socorro-infra/pull/47/files   is
                     that the the whole thing ?  not missing anything ? :)
09:59:56 < rhelmer> phrawzty: heh for starters, that's enough for collection+crashmoving
                    to work, based on our current configs
10:00:13 < rhelmer> phrawzty: it doesn't resolve the whole bug, the bits about not using
                    envconsul for collector etc, but that can wait
```

Based on that, `r+`.

PR #48, opened 2015-03-27
fix bug 1148498; wrapper and split roles

phrawzty:
ping @rhelmer

### DIFF
--- a/terraform/wrapper.sh
+++ b/terraform/wrapper.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+# This script should be run as a wrapper to Terraform.
+# It should be symlinked in consul/ as well.
+
+TFVARS="terraform.tfvars"
+
+# Bail on errors and be verbose.
+set -ex
+
+# Need to specify the environment.
+if [ "${1}x" == "x" ]; then
+    echo "USAGE: ${0} <environment>"
+    exit 1
+fi
+
+# The secret_bucket is specified here.
+if [ ! -r $TFVARS ]; then
+    echo "ERROR: Could not read ${TFVARS}"
+    exit 1
+fi
+
+BUCKET=$(grep secret_bucket terraform.tfvars | awk -F\" '{print $2}')
+ENV=$1
+
+# Grab tfstate from S3
+aws s3 sync --exclude="*" --include="terraform.tfstate" s3://${BUCKET}/tfstate/${ENV}/ ./
+
+# Run TF
+terraform apply -var "environment=${ENV}"
+
+# Upload tfstate to S3
+aws s3 sync --exclude="*" --include="terraform.tfstate" ./ s3://${BUCKET}/tfstate/${ENV}/
+
--- a/terraform/wrapper.sh
+++ b/terraform/wrapper.sh
@@ -1,33 +1,70 @@
 #!/usr/bin/env bash
-# This script should be run as a wrapper to Terraform.
-# It should be symlinked in consul/ as well.
+# This script acts as a wrapper to Terraform.
 
 TFVARS="terraform.tfvars"
+ROLES=(admin buildbox collector consul elasticsearch postgres processor rabbitmq symbolapi webapp)
+HELPARGS=("help" "-help" "--help" "-h" "-?")
 
-# Bail on errors and be verbose.
-set -ex
+function help {
+    echo "USAGE: ${0} <environment> <role>"
+    echo -n "Valid roles are: "
+    local i
+    for i in "${ROLES[@]}"; do
+        echo -n "$i "
+    done
+    echo ""
+    exit 1
+}
 
-# Need to specify the environment.
-if [ "${1}x" == "x" ]; then
-    echo "USAGE: ${0} <environment>"
+function contains_element () {
+    local i
+    for i in "${@:2}"; do
+        [[ "$i" == "$1" ]] && return 0
+    done
+    return 1
+}
+
+# Is this a cry for help?
+contains_element $1 "${HELPARGS[@]}"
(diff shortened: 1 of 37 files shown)


terraform/wrapper.sh
```diff
@@ -0,0 +1,33 @@
...
+fi
+
+# The secret_bucket is specified here.
+if [ ! -r $TFVARS ]; then
+    echo "ERROR: Could not read ${TFVARS}"
+    exit 1
+fi
+
+BUCKET=$(grep secret_bucket terraform.tfvars | awk -F\" '{print $2}')
+ENV=$1
```
rhelmer: maybe we should take another arg here like:
```
INFRA=$3
```
That could be ```.``` by default, or you could pass an arg like ```./wrapper.sh env ./consul```

terraform/wrapper.sh
```diff
@@ -0,0 +1,33 @@
...
+
```
rhelmer: If we go with the ```$INFRA``` idea above, could do a ```pushd $INFRA``` here

terraform/wrapper.sh
```diff
@@ -0,0 +1,33 @@
...
+# This script should be run as a wrapper to Terraform.
+# It should be symlinked in consul/ as well.
+
+TFVARS="terraform.tfvars"
+
+# Bail on errors and be verbose.
+set -ex
+
+# Need to specify the environment.
+if [ "${1}x" == "x" ]; then
```
rhelmer: maybe ```[ $# != 2 ]``` (or whatever expected number of args are)

terraform/wrapper.sh
rhelmer: Also we don't need to do this now, but it'd behoove us to break the ```main.tf``` up further into subdirectories, as we have done with ```./consul``` - ```collector```, ```elasticsearch```, ```webapp``` and so on, and not have a ```./terraform/main.tf``` in here at all.

phrawzty:
This requires that we split out the roles into distinct TF units (as we've discussed previously), so while this script is good, it doesn't *currently* do anything useful. :smile:

phrawzty:
Ok, this splits out all of the roles, which comes with the ugly side effect of creating a tonne of (nearly) duplicate resources all over the goddamn place. :ghost: The wrapper is good for now, but I suspect that we might end up replacing it with a Makefile or something at some point, heh.

@rhelmer `r?`

rhelmer:
The ```README.md``` should mention these changes

rhelmer:
Should there be a symlink in each dir too for ```terraform.tfvars``` ?

phrawzty:
> The README.md should mention these changes

Agreed.

> Should there be a symlink in each dir too for terraform.tfvars ?

`terraform.tfvars` is in `.gitignore`, so that's not possible.  We could add that to the wrapper (now you see why I'm already envisioning a Makefile, hehe).

rhelmer:
> terraform.tfvars is in .gitignore, so that's not possible. We could add that to the wrapper (now you see why I'm already envisioning a Makefile, hehe).

Ah, yeah I think it's starting to make a lot of sense.

terraform/wrapper.sh
```diff
@@ -0,0 +1,70 @@
...
+# Grab tfstate from S3
+aws s3 sync --exclude="*" --include="terraform.tfstate" s3://${BUCKET}/tfstate/${ENV}/ ./
+
+# Run TF
+pushd $ROLE
+terraform apply -var "environment=${ENV}"
+popd
+
+# Upload tfstate to S3
+aws s3 sync --exclude="*" --include="terraform.tfstate" ./ s3://${BUCKET}/tfstate/${ENV}/
```
rhelmer: Do we care about updating this after ```terraform destroy```?

phrawzty: That's a good point; yes, we probably do.  So I guess we're now talking about:
```bash
    echo "USAGE: ${0} <action> <environment> <role>"
```

terraform/wrapper.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# Pre-flight check is good, let's continue.
+ENV=$1
+ROLE=$2
+
+# Be verbose and bail on errors.
+set -ex
+
+# Grab tfstate from S3
+aws s3 sync --exclude="*" --include="terraform.tfstate" s3://${BUCKET}/tfstate/${ENV}/ ./
```
rhelmer: doesn't this need to happen inside the ```pushd``` below? Can there be one ```terraform.tfstate``` for everything?

phrawzty: Oops, yes, sorry, I fixed that but forgot to commit. :blush:

terraform/wrapper.sh
```diff
@@ -0,0 +1,70 @@
...
+# Pre-flight check is good, let's continue.
+ENV=$1
+ROLE=$2
+
+# Be verbose and bail on errors.
+set -ex
+
+# Grab tfstate from S3
+pushd $ROLE
+aws s3 sync --exclude="*" --include="terraform.tfstate" s3://${BUCKET}/tfstate/${ENV}/ ./
```
rhelmer: Do we need to have ```${ROLE}``` in the S3 path too?

terraform/wrapper.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# Run TF
+terraform apply -var "environment=${ENV}"
+
+# Upload tfstate to S3
+aws s3 sync --exclude="*" --include="terraform.tfstate" ./ s3://${BUCKET}/tfstate/${ENV}/
```
rhelmer: Ditto above

rhelmer:
r+

PR #49, opened 2015-03-27
script to assist bootstrapping infra

rhelmer:
Socorro needs to be bootstrapped before it'll come up:

* consul servers need to be joined together
* config needs to be loaded into consul
* PG, ES, webapp, admin/crontabber need to be initialized (```setup-socorro.sh``` can do this)

However there are some other things that should be automatic that aren't yet, which I've made comments about. Generating a new RPM will fix some of them, with the exception of https://github.com/mozilla/socorro/pull/2695

### DIFF
--- a/terraform/bootstrap.sh
+++ b/terraform/bootstrap.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+
+if [ "$1" == "" ]; then
+    echo "Usage: $0 <app>"
+    echo "valid app names are: consul, elasticsearch, postgres, rabbitmq, webapp"
+    exit 1
+fi
+
+app=$1
+
+case $app in
+
+consul)
+  # FIXME download socorro-config backup from S3?
+  tar zxf socorro-config.tar.gz
+  pushd socorro-config
+  ./bulk_load.sh
+  popd
+
+elasticsearch)
+  sudo setup-socorro.sh elasticsearch
+  ;;
+
+postgres)
+  sudo /usr/pgsql-9.3/bin/postgresql93-setup initdb
+  echo FIXME, add listen_addresses = '*' to
+  echo in: /var/lib/pgsql/9.3/data/postgresql.conf
+  echo FIXME, set "md5" for 172.*
+  echo in: /var/lib/pgsql/9.3/data/pg_hba.conf 
+  sudo systemctl restart postgresql-9.3
+  sudo setup-socorro.sh postgres
+  echo FIXME change the password for the breakpad_rw role:
+  echo psql -c "ALTER ROLE breakpad_rw WITH PASSWORD '...';"
+  ;;
+
+rabbitmq)
+  if [ -n "$RABBIT_PASSWORD" ]; then
+      echo "please set the RABBIT_PASSWORD env var"
+      exit 1
+  fi
+  sudo rabbitmqctl add_user socorro ${RABBIT_PASSWORD}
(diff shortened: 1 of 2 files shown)


rhelmer:
Note that there are sufficiently manual steps in the main ```bootstrap.sh``` that they just print the commands rather than trying to actually run them... we can fix this to be more like ```consul/bootstrap.sh``` as we resolve the ```FIXME```s and ```TODO```s

The good news is, I am able to spin up a fully distributed Socorro setup on top of the ```terraform```-created machines if I follow these instructions \o/

terraform/consul/bootstrap.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+if [ $# != 2 ]; then
+    echo "Usage: $0 [public_ips] [private_ips]"
+    exit 1
+fi
+
+PUBLIC_IPS=$1
+PRIVATE_IPS=$1
```
phrawzty: Don't you mean `$2`?

terraform/consul/bootstrap.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+if [ $# != 2 ]; then
+    echo "Usage: $0 [public_ips] [private_ips]"
+    exit 1
+fi
+
+PUBLIC_IPS=$1
+PRIVATE_IPS=$2
```
phrawzty: Don't you mean `$2`?

terraform/consul/bootstrap.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+if [ $# != 2 ]; then
+    echo "Usage: $0 [public_ips] [private_ips]"
```
phrawzty:
```bash
    # It's important to be explicit about the double-quotes here since they *really* matter.
    echo "USAGE: $0 \"<public_ip> [public_ip]\" \"<private_ip> [private_ip]\""

File: terraform/bootstrap.sh [hunk 4]
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
+
+if [ "$1" == "" ]; then
+    echo "Usage: $0 <app>"
+    echo "valid app names are: crontabber, collector, elasticsearch, postgresqlm processor, rabbitmq, webapp"
```
phrawzty: `s/postgresqlm/postgresql,/`

That said, we call it postgres (no ql) elsewhere though - we should try to be consistent.  Also, these line up with "roles" - again, we should try to be consistent with our language.  Thoughts?

phrawzty: It's also possible to be programmatic with the role list (and do some sanity checking on it) as I've done in `wrapper.sh` in #48. Not necessary strictly speaking, but it's a nice touch. :wink:

rhelmer:
@phrawzty @jdotpz is this approach still useful at all, or should we just document how to stand up a brand-spanking-new Socorro cluster?

There are always going to be some things like initializing ES/PG, setting rabbit user/pass, getting consul servers talking to each other etc.

phrawzty:
I rather like the idea of a bootstrapping script, but it does constitute debt: we have to keep it up to date along with everything else.  Given how we lag on doc, I'm not convinced we'd keep both the doc *and* the script up to date, either - perhaps this would be best re-explored down the line, once we've stabilised the setup and config processes further?

PR #50, opened 2015-03-28
fix rabbitmq elb

rhelmer:
r? @phrawzty - we're using the wrong port for the rabbitmq ELB, and it shouldn't be open to the public.

However, note that the ES ELB seems to be open to the world and isn't part of the ```internet_to_elb__http``` security group though so maybe there's more needed there.


phrawzty:
`r+` - let's walk through the config with a fine-toothed comb and sort it out in another PR.

PR #52, opened 2015-04-08
bug 1152427 - use internal ELBs for ES+Rabbit, upgrade to tform 0.4

rhelmer:
@phrawzty, I spent some time today getting our config working in Terraform 0.4 and learned an interesting thing - our account only supports VPC EC2 infra and not Classic. You can see this by going here:

https://console.aws.amazon.com/ec2/v2/home?region=us-west-2

Look at "Supported Platforms" on the right, and "Default VPC" below that. That's why we're getting a default VPC and subnets created for us, and our infra is going into them. As of Terraform 0.4 this means that we have to start specifying our subnet IDs to our AutoScaling Groups.

Thanks to @phinze for finding this - this looked like a different bug at first, and it took a while to figure out that's why we were still hitting the issue.

Incidentally, understanding this helped me to see how we can configure internal ELBs - I've tested at least ```Collector``` and ```Elasticsearch``` roles and they seem to work!

The only downside I can see so far is that nodes now are in the VPC with no public IP address, so we're going to need to figure out some way to SSH into the nodes (bastion host or VPN.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,11 @@ sudo: no
 language: python
 before_install:
   - gem install puppet puppet-lint
-  - pushd terraform; wget 'https://dl.bintray.com/mitchellh/terraform/terraform_0.3.7_linux_amd64.zip'; unzip -u terraform_0.3.7_linux_amd64.zip; popd
+  - pushd terraform; wget 'https://dl.bintray.com/mitchellh/terraform/terraform_0.4.0_linux_amd64.zip'; unzip -u terraform_0.4.0_linux_amd64.zip; popd
 
 install: false
 
 script:
   - puppet-lint --with-filename --no-80chars-check --no-autoloader_layout-check --fail-on-warnings puppet/
   - puppet parser validate `find puppet/ -name '*.pp'`
-  - pushd terraform; ./terraform plan -var="environment=blah" -var="secret_key=FAKE" -var="access_key=FAKE"; popd
-  - pushd terraform/consul; ../terraform plan -var="environment=blah" -var="secret_key=FAKE" -var="access_key=FAKE"; popd
+  - find ./terraform -type d -maxdepth 1 | while read role; do pushd $role; ../terraform plan -var="environment=FAKE" -var="secret_key=FAKE" -var="access_key=FAKE" -var="subnets=FAKE" -var="secret_bucket=FAKE"; popd; done
--- a/terraform/admin/main.tf
+++ b/terraform/admin/main.tf
@@ -4,22 +4,6 @@ provider "aws" {
     secret_key = "${var.secret_key}"
 }
 
-resource "aws_security_group" "private_to_admin__icmp" {
-    name = "${var.environment}__private_to_admin__icmp"
-    description = "Allow pings from within the VPC."
-    ingress {
-        from_port = "-1"
-        to_port = "-1"
-        protocol = "icmp"
-        cidr_blocks = [
-            "172.31.0.0/16"
-        ]
-    }
-    tags {
-        Environment = "${var.environment}"
-    }
-}
-
 resource "aws_security_group" "any_to_admin__ssh" {
     name = "${var.environment}__any_to_admin__ssh"
     description = "Allow (alt) SSH to the Admin node."
@@ -43,15 +27,16 @@ resource "aws_launch_configuration" "lc_for_admin_asg" {
     image_id = "${lookup(var.base_ami, var.region)}"
     instance_type = "t2.micro"
     key_name = "${lookup(var.ssh_key_name, var.region)}"
-    security_groups = [
-        "${aws_security_group.any_to_admin__ssh.name}",
-        "${aws_security_group.private_to_admin__icmp.name}"
-    ]
     iam_instance_profile = "generic"
+    associate_public_ip_address = true
+    security_groups = [
+        "${aws_security_group.any_to_admin__ssh.id}"
+    ]
 }
 
 resource "aws_autoscaling_group" "asg_for_admin" {
     name = "${var.environment}__asg_for_admin"
+    vpc_zone_identifier = ["${split(",", var.subnets)}"]
     availability_zones = [
         "${var.region}a",
         "${var.region}b",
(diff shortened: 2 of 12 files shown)


terraform/admin/main.tf
```diff
@@ -4,22 +4,6 @@ provider "aws" {
...
-        protocol = "icmp"
-        cidr_blocks = [
-            "172.31.0.0/16"
-        ]
-    }
-    tags {
-        Environment = "${var.environment}"
-    }
-}
-
```
rhelmer: I removed the ICMP security groups since we can't use ```cidr_blocks``` this way when specifying a subnet, and they aren't being consistently applied in most of our configs. Even if they were, I don't personally think ICMP is that useful but we could figure it out and reintroduce it if you have strong feelings :)

terraform/admin/main.tf
```diff
@@ -43,15 +27,12 @@ resource "aws_launch_configuration" "lc_for_admin_asg" {
     image_id = "${lookup(var.base_ami, var.region)}"
     instance_type = "t2.micro"
     key_name = "${lookup(var.ssh_key_name, var.region)}"
-    security_groups = [
-        "${aws_security_group.any_to_admin__ssh.name}",
-        "${aws_security_group.private_to_admin__icmp.name}"
-    ]
```
rhelmer: If we leave ```security_groups``` in the LC, then ```terraform apply``` will be fine but when the ASG tries to launch an instance it'll fail with:
```
"Launching a new EC2 instance. Status Reason: The 
parameter SecurityGroup is not recognized. Launching EC2 
instance failed."
```

Since they are going into the VPC now and aren't publicly routable I don't think this is useful.

rhelmer:
>  The only downside I can see so far is that nodes now are in the VPC with no public IP address, so we're going to need to figure out some way to SSH into the nodes (bastion host or VPN.)

There might be some way to give the nodes a public IP address, but I wonder if it isn't better just to go through the pain of setting up a bastion host or VPN now, since it seems like it reduces our surface area for attacks.

rhelmer:
> There might be some way to give the nodes a public IP address, but I wonder if it isn't better just to go through the pain of setting up a bastion host or VPN now, since it seems like it reduces our surface area for attacks.

Turns out we *can* set the subnet (already done by default) and the LCs to allow public IP addresses for the nodes - I've done this in the PR, and adjusted the security groups so they work.

I've at least tested that I can spin up consul and connect to it from the collector nodes. I have not tested that ELBs work right yet (internal or otherwise.)

@phrawzty this is ready for review now.

rhelmer:
Oh, and our travis config needs to be updated - it is only running ```terraform plan``` in the ```./terraform``` and ```./terraform/consul```. So don't trust the green travis result in this PR, it's only testing the consul dir :)

terraform/admin/main.tf
```diff
@@ -43,15 +27,16 @@ resource "aws_launch_configuration" "lc_for_admin_asg" {
     image_id = "${lookup(var.base_ami, var.region)}"
     instance_type = "t2.micro"
     key_name = "${lookup(var.ssh_key_name, var.region)}"
+    iam_instance_profile = "generic"
+    associate_public_ip_address = true
```
rhelmer: This is how we get the instances the ASG spins up to have public IP addresses.

terraform/admin/main.tf
```diff
@@ -43,15 +27,16 @@ resource "aws_launch_configuration" "lc_for_admin_asg" {
...
     security_groups = [
-        "${aws_security_group.any_to_admin__ssh.name}",
-        "${aws_security_group.private_to_admin__icmp.name}"
+        "${aws_security_group.any_to_admin__ssh.id}"
```
rhelmer: If you're launching EC2 "classic" infra you pass the name, for EC2 "VPC" infra you must use the id, otherwise everything will launch but ASG will fail to spin up instances and complain about invalid parameter ```SecurityGroup```.

This PR makes us use VPC everywhere to avoid future confusion.

.travis.yml
```diff
@@ -2,12 +2,11 @@ sudo: no
...
+  - pushd terraform; wget 'https://dl.bintray.com/mitchellh/terraform/terraform_0.4.0_linux_amd64.zip'; unzip -u terraform_0.4.0_linux_amd64.zip; popd
 
 install: false
 
 script:
   - puppet-lint --with-filename --no-80chars-check --no-autoloader_layout-check --fail-on-warnings puppet/
   - puppet parser validate `find puppet/ -name '*.pp'`
-  - pushd terraform; ./terraform plan -var="environment=blah" -var="secret_key=FAKE" -var="access_key=FAKE"; popd
-  - pushd terraform/consul; ../terraform plan -var="environment=blah" -var="secret_key=FAKE" -var="access_key=FAKE"; popd
+  - for role in admin buildbox collector consul elasticsearch postgres processor rabbitmq symbolapi webapp; do pushd $role; ./terraform plan -var="environment=blah" -var="secret_key=FAKE" -var="access_key=FAKE"; popd; done
```
phrawzty:
```
for i in `ls`; do pushd $i; terraform plan; [...]
```

:question:

terraform/postgres/main.tf
```diff
@@ -59,10 +43,9 @@ resource "aws_instance" "postgres" {
     count = 1
     security_groups = [
         "${aws_security_group.private_to_postgres__postgres.name}",
-        "${aws_security_group.any_to_postgres__ssh.name}",
-        "${aws_security_group.private_to_postgres__icmp.name}"
+        "${aws_security_group.any_to_postgres__ssh.name}"
     ]
-    block_device {
+    ebs_block_device {
```
rhelmer: This changed in tform 0.4

terraform/admin/main.tf
phrawzty: Yeah that's cool. We can ICMP later if necessary.

ICP

phrawzty:
Regarding VPCs: we get 5 total. I envision using one VPC per environment (so one each for stage and prod), with any number of required subnets in each, respectively.

phrawzty:
```
vpc_zone_identifier = ["${split(",", var.subnets)}"]
```

This is good. We need to document it in the README since it is necessary to make things works (i.e. `tfvars`).

phrawzty:
`r+` pending README and failing Travis. :grin:

PR #55, opened 2015-04-15
fix bug 1152617 - auto-join consul cluster

rhelmer:


rhelmer:
@phrawzty - I took a quick stab at this, per comments in the bug:
* pass in domain name as terraform variable, for puppet to use
* put an ELB in front of consul's RPC service
* assume there is a ```consul.$::socoro_domain``` DNS entry pointing to above ELB

This means that bringing up a new cluster from scratch becomes:

1) bring up consul cluster first and bootstrap it
2) make DNS entry for consul ELB
3) bring up the rest of the roles, which will now auto-join the consul cluster

Don't merge this yet as it's totally untested :) I think it's probably on the right track though, lmk what you think.

rhelmer:
I wonder if it's be clearer to have a variable like ```consul_host``` rather than just ```domain``` that we assume has a ```consul``` subdomain...

puppet/modules/socorro/manifests/role/admin.pp
```diff
@@ -3,16 +3,21 @@
...
   file {
     '/etc/cron.d/socorro':
-      mode   => '0600',
-      owner  => root,
-      group  => root,
-      source => 'puppet:///modules/socorro/etc_cron.d/socorro'
+      mode    => '0600',
+      owner   => root,
+      group   => root,
+      source  => 'puppet:///modules/socorro/etc_cron.d/socorro',
```
jdotpz: Should we throw a Notify in here to tell that any box built for the first time should join the consul cluster?

rhelmer:
OK I've tested this with collector at least, and it seems to work. @phrawzty would you mind reviewing this?

phrawzty:
As we discussed on IRC, I'm not sure that passing this to `user_data` via `tfvar` is the way to go, especially since Consul is itself managed by Terraform. I'm also concerned about that script becoming a catch-all crutch for things which might be managed in other ways.

I propose populating the S3-based Hiera bucket with a "consul_hostname" (or whatever) so that we can let Puppet do the work for us.  From there, we can create a `role/init.pp` which calls calls `hiera()`, sets a variable, and then runs an `exec` to join the cluster.  We would then just `include socorro::role` in each of the roles.

rhelmer:
@phrawzty OK I reverted my changes that passed this in via ```user_data```, the only concern I have now is - how do different environments set ```consul_hostname```?

puppet/modules/socorro/manifests/role/admin.pp
```diff
@@ -12,7 +12,7 @@
       owner   => root,
       group   => root,
       source  => 'puppet:///modules/socorro/etc_cron.d/socorro',
-      require => Package['socorro']
+      require => [Package['socorro'], Exec['join_consul_cluster']]
```
phrawzty: Style nit - should be:
```puppet
require => [
  Package['socorro'],
  Exec['join_consul_cluster']
]
```

Repeat for all further instances thereof.

puppet/modules/socorro/manifests/role/admin.pp
```diff
@@ -12,7 +12,10 @@
       owner   => root,
       group   => root,
       source  => 'puppet:///modules/socorro/etc_cron.d/socorro',
-      require => Package['socorro']
+      require => [
```
phrawzty: You're going to want to `include socorro::role` at the top of this manifest so that the `Exec[]` below will be accessible.  Same for all instances below.

phrawzty:
At long last!  Squash then `r+`.

PR #56, opened 2015-04-15
 bug 1151911 - add SSL for collector+webapp

rhelmer:


terraform/variables.tf
```diff
@@ -3,6 +3,8 @@ variable "access_key" {}
 variable "secret_key" {}
 variable "secret_bucket" {}
 variable "subnets" {}
+variable "collector_cert" {}
```
phrawzty: Are `collector_cert` and `webapp_cert` in the secret bucket?

terraform/variables.tf
```diff
@@ -3,6 +3,8 @@ variable "access_key" {}
...
+variable "webapp_cert" {}
```
phrawzty: These files (or wildcards?) should be in `.gitignore` as well.

terraform/variables.tf
rhelmer: They are simply names of the certs loaded in AWS, the terraform docs example is ```arn:aws:iam::123456789012:server-certificate/certName```

terraform/variables.tf
rhelmer: @phrawzty the cert itself lives in AWS so neither the terraform user nor the instances need to have the files themselves locally.

phrawzty: Aaah ok that's great.

phrawzty:
`r+`

PR #57, opened 2015-04-16
Add app and project tags to terraform configs

jdotpz:


rhelmer:
I'm not sure if we should use "crashreporter" for the project... we've used "socorro" historically (which is a codename I guess so kind of weird), I think people might confuse "crashreporter " for "crash-reports" aka "collector".

jdotpz:
+1, updated that projectname.

rhelmer:
r+++ would review again

rhelmer:
@jdotpz want to squash this down to one commit? Also do you need me to merge or are you able to do that?

phrawzty:
Before we merge I have a nit, heh.  I propose `s/app/role/` which is more in line with the established nomenclature of the rest of the project.

jdotpz:
Sure, that's sensible.

phrawzty:
Squash then `r+`.

rhelmer:
This is done as part of #64 instead.

PR #59, opened 2015-04-16
bump processor instance_type to one with more RAM

rhelmer:
r? @phrawzty / @jdotpz - the OOM killer kicks in on an ```t2.micro``` processing even a single job :+1: 

I think ```t2.medium``` should at least work at all, and we can have it run a couple threads.

### DIFF
--- a/terraform/processor/main.tf
+++ b/terraform/processor/main.tf
@@ -24,7 +24,7 @@ resource "aws_launch_configuration" "lc_for_processor_asg" {
     name = "${var.environment}__lc_for_processor_asg"
     user_data = "${file(\"socorro_role.sh\")} ${var.puppet_archive} processor ${var.secret_bucket} ${var.environment}"
     image_id = "${lookup(var.base_ami, var.region)}"
-    instance_type = "t2.micro"
+    instance_type = "r3.large"
     key_name = "${lookup(var.ssh_key_name, var.region)}"
     iam_instance_profile = "generic"
     associate_public_ip_address = true


jdotpz:
Perhaps we can test out a m3.medium.

screenshot 2015-04-15 23 58 53
If we're hitting that much CPU doing that little, we'll burn through our CPU credits on the t2 class.

rhelmer:
@jdotpz ok makes sense thanks - am I reading http://aws.amazon.com/ec2/instance-types/#m3 right though does the m3.medium have just a 4GB SSD?

The processors manage a symbol cache (in /tmp/symbols by default) which we allow to grow to 30GB which is enough to get over 90% hit rate when @luser looked into it recently.

So we might want to go to the next one up. We should be able to use the extra cores and RAM.

rhelmer:
For reference, in PHX we have 10 processor servers with 32GB RAM and 4-core Xeons: http://ark.intel.com/products/47923/Intel-Xeon-Processor-E5640-12M-Cache-2_66-GHz-5_86-GTs-Intel-QPI

Here're some numbers from the beginning of the year until today from Socorro's own little processor stats table (```server_status```):

```
processor jobs per second

 min |       avg        |   max   |      stddev      
-----+------------------+---------+------------------
   0 | 5.34128255849546 | 79.5384 | 1.83588421555469
```

It looks like the ganglia graphs in PHX aren't collecting new data anymore, but I can see data over the past year like CPU and RAM and network - there is an obvious growing trend as we migrated data to S3, but overall memory use, load average and CPU are all very low across the servers in PHX.

Really we have so many in PHX to be able to handle spikes, which we can handle using autoscaling in AWS.

phrawzty:
Recall that we put `t2.micro` only because it's the smallest possible instance type - it has always been the plan to use, uh, more appropriate instance types for, you know, _real life_. :grinning: 

> The processors manage a symbol cache (in /tmp/symbols by default) which we allow to grow to 30GB which is enough to get over 90% hit rate when @luser looked into it recently. So we might want to go to the next one up. We should be able to use the extra cores and RAM.

`m3.large` is 1 x 32 GB; if we want to be able to use up to 30 GB per node, that'd be cutting awfully damn close.  Can we use EBS for this or do we want it to be entirely ephemeral?  In the latter case we'd want a larger cushion, the next step up being `m3.xlarge` at 2 x 40 GB.

rhelmer:
```m3.large``` sgtm. I think currently we can't touch the launch config without destroying the ASG too, but if we configure terraform to ```lifecycle { create_before_destroy = true }``` and to not have a name (to avoid duplication) then we should be able to fix this.

Fortunately this doesn't really matter much for processor, since they don't have an ELB in front of it - this is going to be an issue for the rest of our infra though.

jdotpz:
*I think currently we can't touch the launch config without destroying the ASG too,*
Nope, cleared to touch it in all of the ways.  You just can't edit ELBs associated with the group, and sans an ELB, we can change all of the things.

Will that disk be holding anything cache-y(swap or otherwise) or otherwise in need of fast storage?  If so, EBS and EBS optimized all the way for sure.

rhelmer:
>>I think currently we can't touch the launch config without destroying the ASG too,
>Nope, cleared to touch it in all of the ways. You just can't edit ELBs associated with the group, and sans an ELB, we can change all of the things.

Oh yeah we can do it, sorry I meant terraform won't do it for us due to our current config as I recently learned in #terraform-tool on freenode - IIRC we have to specify ```create_before_destroy``` on the resource and also the LC must not have a name, since terraform isn't smart enough yet to make up a unique one. We can tell what it using using tags, so maybe no big deal?

And actually, I just ran ```terraform plan``` and it's offering to do the right thing: modify (~) aws_autoscaling_group.asg_for_processor and destroy/create (-/+) aws_launch_configuration.lc_for_processor_asg - so maybe I am wrong and it'll handle it? We'll have to try it :)

> Will that disk be holding anything cache-y(swap or otherwise) or otherwise in need of fast storage? If so, EBS and EBS optimized all the way for sure. 

The only thing we need disk for is cache - the symbols are downloaded from S3 as-needed and cached in ```/tmp/symbols``` on the processors. The processors themselves take care of cleanup (they maintain an LRU cache)

Would it be better to use instance storage or EBS for that?

jdotpz:
EBS for sure.  SSD EBS is dirt cheap and much more predictable and consistent than instance storage.  

The one downside is usually when there's a big honkin' problem on AWS, it's EBS that's left a stain on the bedding.

jdotpz:
We'll have to bust out our calculators and see if running a fleet of oversized (cpu and mem-wise) servers to get the included SSD sizing outweighs the costs of just having our base AMI include that storage for a few additional of Uncle Lonnen's dollars.

rhelmer:
OK thanks. BTW filed bug 1155524 on the terraform launch-config bit, that's really a separate issue.

rhelmer:
> We'll have to bust out our calculators and see if running a fleet of oversized (cpu and mem-wise) servers to get the included SSD sizing outweighs the costs of just having our base AMI include that storage for a few additional of Uncle Lonnen's dollars.

Hm yeah that is a pretty big price difference, I'm also not sure yet what kind of CPU+RAM we're going to need t handle normal workload - is it going to be generally cheaper to have one server that can handle the full workload (and scale more if we have a spike), or a bunch of lower-powered instances to spread the work to? Or does it just depend on the vagaries of AWS pricing and the exact resources we need?

jdotpz:
There's some inherent good to having at least a couple servers up so you can be spread across AZ's so when one aforementioned stain in the bed appears in one AZ, you are left scrubbing it out.

That said, it depends on how we want to scale, and the peculiarities of our apps being hosted.  

Bumping up horizontally instead of starting off vertically can be great if we scale super hard on first indication that we have to scale.  Since we'll be using AMI's, our time to go from "crap, CPU is too high for our existing cluster, scale up!" to "nodes ready for service" will be < ~45 seconds.  However, then we'll have to figure what to do with the cooloff period timing.  That's usually around 5 minutes (to let the panicking cpus settle down and not cause a cascade of scaling).  For that reason, it's usually proper to scale super hard, super fast...
Say you have it set to 50% CPU aggregate on the cluster causes a scale up.  3 availability zones, so you multiply the scaleup # by that.  Scale up say, 24 servers on one alarm to go up.  Then, each "alarm" for low cpu to trigger a scale down brings it down by just 3.  The right idea is to scale fast and hard on the way up, and casual on the way down.

jdotpz:
Interesting side note to scaling:
You can scale based on any metric or alarm in Cloudwatch.  And, it turns out, you can send your own custom metrics up to Cloudwatch.  We could scaleup and down based on how many animated gifs Lonnen sends in an afternoon.
https://www.loggly.com/blog/send-custom-metrics-to-cloudwatchs-api/

rhelmer:
r? @phrawzty @jdotpz OK looking at the available options, I think `r3.large` give us the best value, since it's got a lot more memory (which is really the limiting factor on how many simultaneous processor jobs we can) and also supports EBS.

It supports enhanced networking, which we should figure out to turn on at some point (cause why not?)

rhelmer:
verbal r+ ftw

PR #60, opened 2015-04-17
fix bug 1155504 - add crash-analysis host

rhelmer:
r? @phrawzty @jdotpz - I think this is the most straightforward config right now for an EC2 instance running just Nginx.

The puppet bit seems pretty reasonable, but we really need to cut down on the amount of terraform config needed :)

### DIFF
--- a/terraform/analysis/main.tf
+++ b/terraform/analysis/main.tf
@@ -1,0 +1,131 @@
+provider "aws" {
+    region = "${var.region}"
+    access_key = "${var.access_key}"
+    secret_key = "${var.secret_key}"
+}
+
+resource "aws_security_group" "any_to_analysis__ssh" {
+    name = "${var.environment}__any_to_analysis__ssh"
+    description = "Allow (alt) SSH to the CrashAnalysis node."
+    ingress {
+        from_port = "${var.alt_ssh_port}"
+        to_port = "${var.alt_ssh_port}"
+        protocol = "tcp"
+        cidr_blocks = [
+            "0.0.0.0/0"
+        ]
+    }
+    tags {
+        Environment = "${var.environment}"
+        role = "crash-analysis"
+        project = "socorro"
+    }
+}
+
+resource "aws_security_group" "internet_to_analysis_elb__http" {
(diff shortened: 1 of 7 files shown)


terraform/analysis/main.tf
```diff
@@ -0,0 +1,129 @@
...
+    description = "Allow (alt) SSH to the CrashAnalysispp node."
+    ingress {
+        from_port = "${var.alt_ssh_port}"
+        to_port = "${var.alt_ssh_port}"
+        protocol = "tcp"
+        cidr_blocks = [
+            "0.0.0.0/0"
+        ]
+    }
+    tags {
```
phrawzty: Should the tags here (and elsewhere) also have the `role` and `project` tags (like in #57)?

phrawzty:
What does this role do? The service config is very, very light - could be anything?

phrawzty:
> The puppet bit seems pretty reasonable, but we really need to cut down on the amount of terraform config needed :)

Yeah - I'd really like to make use of templates more (I guess?).

rhelmer:
> What does this role do? The service config is very, very light - could be anything?

The bug has a little more info - crash-analysis just hosts some Apache UserDirs, I expect it'll need a few other things to run server-side reports (PHP at least, I could at that now I guess - note that it only needs the PHP CLI command, nginx doesn't need to run PHP)

rhelmer:
@phrawzty OK I think that addresses your comments, lmk

terraform/analysis/main.tf
```diff
@@ -0,0 +1,131 @@
+provider "aws" {
+    region = "${var.region}"
+    access_key = "${var.access_key}"
+    secret_key = "${var.secret_key}"
+}
+
+resource "aws_security_group" "any_to_analysis__ssh" {
+    name = "${var.environment}__any_to_analysis__ssh"
+    description = "Allow (alt) SSH to the CrashAnalysispp node."
```
phrawzty: Speeling. :grin:

phrawzty:
Squash and then `r+` w/ minor nits.

PR #61, opened 2015-04-17
nginx should not start until socorro package is in place

rhelmer:
r? @jdotpz the Socorro RPM drops some nginx configs into ```/etc/nginx/conf.d```

Eventually we're going to have to override this in a smarter way (to drop the real hostname into ```server_name```) but for now this will get this going.

### DIFF
--- a/puppet/modules/socorro/manifests/role/collector.pp
+++ b/puppet/modules/socorro/manifests/role/collector.pp
@@ -31,7 +31,8 @@ include socorro::role::common
       ensure=> latest;
 
     'nginx':
-      ensure=> latest;
+      ensure=> latest,
+      require => Package['socorro'];
   }
 
 }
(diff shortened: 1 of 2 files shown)


puppet/modules/socorro/manifests/role/collector.pp
```diff
@@ -31,7 +31,8 @@
       ensure=> latest;
 
     'nginx':
-      ensure=> latest;
+      ensure=> latest,
```
phrawzty: Please align this (not sure why `puppet-lint` didn't complain here).

puppet/modules/socorro/manifests/role/webapp.pp
```diff
@@ -31,7 +31,8 @@
       ensure=> latest;
 
     'nginx':
-      ensure=> latest;
+      ensure=> latest,
```
phrawzty: Please align this (not sure why `puppet-lint` didn't complain here).

rhelmer:
So I think we don't actually need this - we're going to change the nginx RPM so it doesn't activate sample configs. We're going to need a PR to actually drop the configs in place, but that's separate from this.

PR #62, opened 2015-04-17
Consul running before join

rhelmer:
Normal nodes need to have consul agent running before trying `consul join`, and consul agent must be running in server mode on consul servers.

### DIFF
--- a/puppet/modules/socorro/manifests/role/consul.pp
+++ b/puppet/modules/socorro/manifests/role/consul.pp
@@ -1,6 +1,6 @@
 # Set up a Consul Server node.
 
-include socorro::role::common
+class socorro::role::consul {
 
   service {
     'consul':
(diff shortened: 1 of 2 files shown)


puppet/modules/socorro/manifests/role/consul.pp
```diff
@@ -27,4 +26,13 @@
       require => Package['consul-ui'];
   }
 
+  # We expect this to come from the secret S3 bucket
+  # This is here instead of common.pp so we can depend on server.json
+  $consul_hostname = hiera("${::environment}/consul_hostname")
+  exec {
+      'join_consul_cluster':
+        command => "/usr/bin/consul join ${consul_hostname}",
+        require => File['/etc/consul/server.json']
```
phrawzty: The service is already subscribed (implied require) to the file so the exec can (should?) require the service. :spaghetti:

rhelmer:
I think there must be a clearer way to do this, and this is not a super important problem to solve right now. Closing for now.

PR #63, opened 2015-04-23
[DO NOT MERGE] generate symlinks in wrapper

phrawzty:
ping @rhelmer @jdotpz 

If this looks good I'll go through the motions to remove the symlinks, then we merge.


terraform/wrapper.sh
```diff
@@ -24,6 +25,14 @@ function contains_element () {
     return 1
 }
 
+function check_symlinks () {
+    for i in ${SYMLINKS[@]}; do
+        if [ ! -h $i ]; then
```
rhelmer: `-h` is deprecated in favor of `-L`

phrawzty:
Verbal `r+`. :wine_glass:

PR #64, opened 2015-04-23
bug 1155524; Rename infrastructure, make a master elb name

jdotpz:
`r?` @phrawzty @rhelmer

PR #65, opened 2015-04-23
bug 1155524 - terraform should be able to make launch config changes wit...

rhelmer:
...hout tearing out infra

r? @phrawzty @jdotpz - as discussed, the only way terraform supports this (per `#terraform-tool` on `irc.freenode.net`) is to:

* set `create_before_destroy = true` in `lifecycle` for the resource
* let terraform name the resource (otherwise the "create" step will fail on AWS due to duplicate name)

Unfortunately, launch configs don't seem to support tags, so the tradeoff here is that we have to look at an ASG in order to tell where a given LC belongs :/

I made sure both `terraform plan` and `terraform apply` works with this patch, for at least the processor role.

### DIFF
--- a/terraform/admin/main.tf
+++ b/terraform/admin/main.tf
@@ -24,7 +24,6 @@ resource "aws_security_group" "ec2-socorroadmin-sg" {
 
 # Admin (crontabber, etc)
 resource "aws_launch_configuration" "lc-socorroadmin" {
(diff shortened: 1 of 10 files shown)


phrawzty:
This will need to be re-based against #64 when that lands.

phrawzty:
R++++++++ would review again

PR #67, opened 2015-04-23
tests need to run with -e so any failed test causes non-0 exit

rhelmer:
I noticed that our tests aren't actually failing the travis job, even though the tests are failing.

The way we have the test shoved into `.travis.yaml` is nigh unreadable, let's split it out.

r? @phrawzty @jdotpz


phrawzty:
# `r+`

PR #68, opened 2015-04-23
add opt to generate symlinks

phrawzty:
Passing `symlinks` to `wrapper.sh` will just generate all the symlinks and then exit.

We can use this in Travis, for example.

`r?` @rhelmer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ terraform.tfstate*
 *.swp
 *.plan
 *.log
+terraform/*/*.sh
+terraform/*/*.tf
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -5,3 +5,5 @@ terraform.tfstate*
 *.swp
 *.plan
 *.log
+terraform/*/*.sh
+terraform/*/*.tf
```
rhelmer: `main.tf` should be allowed :)

rhelmer:
r+

PR #70, opened 2015-04-24
security groups need to be created before destroy too

rhelmer:
r? @jdotpz


jdotpz:
r+, would merge again.

PR #73, opened 2015-04-24
set port for collector sg correctly

rhelmer:
r? @phrawzty @jdotpz


rhelmer:
I tested this both `plan` and `apply` work fine.

jdotpz:
+1

PR #74, opened 2015-04-24
ELB SG needs to be on LC

rhelmer:
r? @jdotpz @phrawzty - this is why the ES ELB on stage isn't going live.

I already applied it on stage, applied cleanly and fixes the problem.


phrawzty:
`r+`

PR #76, opened 2015-04-28
fix bug 1158924, land socorro-config in public repo

rhelmer:


rhelmer:
This was reviewed by @phrawzty and @jdotpz in the bug.

PR #77, opened 2015-04-28
nobug; amplify docs and minor fixes

phrawzty:
Filled out the README for Puppet. As an added bonus I fixed some Puppet syntax and sorting issues, and removed `del_on_term` (which is not used).

@rhelmer `r?`

### DIFF
--- a/puppet/README.md
+++ b/puppet/README.md
@@ -1,3 +1,44 @@
-# Puppet
+# Default manifest
 
-# Details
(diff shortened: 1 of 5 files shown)


rhelmer:
r+ lgtm

phrawzty:
Turns out that `del_on_term` is still used by the Postgres role, so I reverted `variables.tf` for now.

PR #82, opened 2015-04-30
Healthcheck tuning, cross zone elb, missing tags, match elb<->as zones, max_size /\

jdotpz:
* Changes health checks to be more reasonable (will speed up the deploy)
* Adds some missing zones that AS groups scaled into but elbs didn't have them on
* Adds some missing tags
* Turns on cross zone balancing, which might not be what we want long term, but for stage at 1 instance, it might help.
* Makes AS groups max_size 10, to allow for room to scale in new deploys while keeping existing working infra

### DIFF
--- a/terraform/admin/main.tf
+++ b/terraform/admin/main.tf
@@ -53,7 +53,7 @@ resource "aws_autoscaling_group" "as-socorroadmin" {
         "aws_launch_configuration.lc-socorroadmin"
     ]
     launch_configuration = "${aws_launch_configuration.lc-socorroadmin.id}"
-    max_size = 1
+    max_size = 10
     min_size = 1
     desired_capacity = 1
     tag {
--- a/terraform/analysis/main.tf
+++ b/terraform/analysis/main.tf
@@ -64,11 +64,19 @@ resource "aws_elb" "elb-socorroanalysis" {
     security_groups = [
         "${var.elb_master_web_sg_id}"
     ]
+    health_check {
+      healthy_threshold = 2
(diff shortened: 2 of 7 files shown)


terraform/admin/main.tf
```diff
@@ -53,7 +53,7 @@ resource "aws_autoscaling_group" "as-socorroadmin" {
         "aws_launch_configuration.lc-socorroadmin"
     ]
     launch_configuration = "${aws_launch_configuration.lc-socorroadmin.id}"
-    max_size = 1
+    max_size = 10
```
rhelmer: Do we care about tuning this per-app at all? We never want more "admin" nodes running in practice.

terraform/analysis/main.tf
```diff
@@ -64,11 +64,19 @@ resource "aws_elb" "elb-socorroanalysis" {
     security_groups = [
         "${var.elb_master_web_sg_id}"
     ]
+    health_check {
+      healthy_threshold = 2
+      unhealthy_threshold = 2
+      timeout = 3
+      target = "TCP:80/"
+      interval = 12
+    }
```
rhelmer: so much better

rhelmer:
r+ lgtm

terraform/admin/main.tf
jdotpz: It won't ever really autoscale, it'd have to be something we manually did, so it won't hurt anything, but will make it possible to say, bring up a clone node to test something.

PR #84, opened 2015-05-04
Change latest to installed for socorro packages

jdotpz:
This is meant to address this scenario:
* v1.0.1 is in production, v1.0.2 is in staging being tested/updated
* We scale in a new node into a production cluster
* The new production node has v1.0.1 baked in on instantiation
* Puppet runs on the new production node, upping it to v1.0.2 despite that not being cleared for prod

This will mean we will have to use a base AMI (currently used) for the build instead of querying the AWS API for the latest staging AMI.  Otherwise, it would not install the updated version while making an AMI for staging and eventual production.

### DIFF
--- a/puppet/modules/socorro/manifests/role/collector.pp
+++ b/puppet/modules/socorro/manifests/role/collector.pp
@@ -28,7 +28,7 @@ include socorro::role::common
 
   package {
     'socorro':
-      ensure=> latest;
+      ensure=> installed;
 
     'nginx':
       ensure=> latest;
--- a/puppet/modules/socorro/manifests/role/admin.pp
+++ b/puppet/modules/socorro/manifests/role/admin.pp
@@ -5,7 +5,7 @@ include socorro::role::common
 
   package {
     'socorro':
-      ensure => latest
+      ensure => installed
   }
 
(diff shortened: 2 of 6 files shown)


phrawzty:
That's a clever solution to the problem - good thinking!

puppet/modules/socorro/manifests/role/collector.pp
```diff
@@ -28,7 +28,7 @@
 
   package {
     'socorro':
-      ensure=> latest;
+      ensure=> installed;
```
phrawzty: Please put a space between the keyword and the fat arrow, i.e.
```puppet
    ensure => installed
```

Not sure why the linter didn't catch this.

Same same for the other manifests in this PR.

jdotpz:
Rob has a more clever option actually, which is to only do a package install on the packer run, not the base puppet repo.  That seems a bit more ideal.

phrawzty:
Up to you whether you want to fix and merge or just close, then.  I leave it with @rhelmer and you to decide.

rhelmer:
I think `latest` versus `installed` shouldn't make any difference given #86.

PR #85, opened 2015-05-04
Back to secgroup-per-env/app combo

jdotpz:
In retrospect, having one security group shared is less desirable than having the ability to spin up test/clone infrastructure in the same VPC without polluting statefiles with different VPC ids.

I'm running through the test applies for this now.


rhelmer:
r+

PR #86, opened 2015-05-04
bug 1147028 - put puppet into AMI instead of using S3

rhelmer:
r? @phrawzty @jdotpz - as we discussed, this moves us away from making changes to the AMI after generating it, puppet is only used by the user_data in the LC for starting up services for the desired role.

Note I haven't tested to make sure that packer puts the files in the expected place in the format we want, I'd appreciate comment on whether this is an appropriate place to keep manifests too.


jdotpz:
+1 on location, and +1 to testing packer with it.  I can pull that onto the builder box I've got running and test it tonight.

rhelmer:
@phrawzty @jdotpz this PR now also moves all installation to packer, so the roles (run by the launch config) only do things like starting services and dropping config files in.

packer/socorro_base.json
```diff
@@ -24,6 +24,7 @@
         {
             "type": "puppet-masterless",
             "manifest_file": "../puppet/manifests/default.pp",
+            "staging_directory": "/etc/puppet/socorro",
```
phrawzty: What does the final dir structure look like here?  As in, what exactly gets put under `/etc/puppet/socorro/` ?

puppet/modules/socorro/manifests/packer/base.pp
```diff
@@ -46,7 +46,10 @@
       'mod_wsgi',
       'rabbitmq-server',
       'supervisor',
-      'unzip'
+      'unzip',
+      'php-cli',
+      'bind-utils',
+      'consul-ui'
```
phrawzty: Please, help this old sysadmin keep his sanity, and sort these alphabetically. :hamster:

puppet/modules/socorro/manifests/role/analysis.pp
```diff
@@ -6,34 +6,7 @@
...
-      require => File['/etc/nginx/nginx.conf'];
-  }
-
-  package {
-    'nginx':
-      ensure=> latest;
-
-    'php-cli':
-      ensure=> latest;
+      enable  => true;
```
phrawzty: Too much whitespace between the keyword and the fat arrow. Unsure why the linter didn't catch this.

terraform/socorro_role.sh
```diff
@@ -1,26 +1,13 @@
...
 
-DIR="/tmp/${RANDOM}-${RANDOM}"
-
 function socorro_role {
-    # Set up the working dir.
-    mkdir $DIR
-    pushd $DIR
-
-    # Provide the secret bucket name to Hiera (hiera-s3).
-    sed -i "s:@@@SECRET_BUCKET@@@:${3}:" /etc/puppet/hiera.yaml
```
phrawzty: You removed this line, which is fine, but how does Hiera become aware of the secret bucket name now? I don't see what in this PR replaces this functionality.

phrawzty:
The various `main.tf`'s still pass `${var.puppet_archive}` via userdata, and `terraform/socorro_role.sh` still expects an equivalent positional argument. These will need to be adjusted as well.

jdotpz:
We need to land this before I can add prod infra and test updates to stage vs. prod in the deploy script.

terraform/socorro_role.sh
rhelmer: Yeah we don't want to bake secrets into the AMI since we make them public right? I'll restore this.

packer/socorro_base.json
rhelmer: I think it's just the puppet files, I will test today.

rhelmer: To expand a bit, I'd expect `socorro-infra/puppet/modules/socorro/` to go to `/etc/puppet/socorro` in the AMI and be left there by packer, with this change. This is what the user_data script is expecting.

phrawzty: In this case, the target should follow the same hierarchy as the source, so `/etc/puppet/modules/socorro/`.

rhelmer:
This has been merged into PR #95, going to let it land as part of that.

PR #87, opened 2015-05-05
do not complain about quoted args like: ./wrapper.sh "apply -var a=b" en...

rhelmer:
...v role"

### DIFF
--- a/terraform/wrapper.sh
+++ b/terraform/wrapper.sh
@@ -51,7 +51,7 @@ if [ "${1}x" == "x" ]; then
 fi
 
 # Did we want to generate symlinks?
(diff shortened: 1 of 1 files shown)


jdotpz:
r+

phrawzty:
This is the sort of stuff a bash linter might be able to catch. I suppose we could integrate the downloadable version of [ShellCheck](http://www.shellcheck.net/) into our CI flow?

In fact, running `wrapper.sh` through the checker reveals a number of potential double-quoting issues, heh.

rhelmer:
@phrawzty that sounds great - I think we should definitely use a shell linter, I wasn't aware that there was a decent one out there.

PR #88, opened 2015-05-05
re-add missing SG

rhelmer:
r? @jdotpz @phrawzty - this SG went missing, so the ELB couldn't reach our instance and would never go in service.

I rearranged things a little bit to make diffing against the ES `main.tf` easier.


rhelmer:
BTW I tested this and applied against stage already, since stage was busted.

jdotpz:
(Echoing irc discussion)
We probably just need to add an ingress rule for the rmq port from elb->ec2 sg, ala https://www.irccloud.com/pastebin/RbfiabCM

rhelmer:
@jdotpz wfm, I am just aping what ES does, so we should fix it in both places.

jdotpz:
Here's the fix for ES I think.
https://www.irccloud.com/pastebin/jRgmERii

PR #90, opened 2015-05-05
Use variables for most autoscale group sizing and instance types

jdotpz:
# Control instance sizing and scaling with variables

1) How do we safely scale in new nodes in a dynamic autoscaling group which may be scaled up, being sure not to knock out a bunch of nodes we need during a terraform apply?

2) How do we write terraform scripts accounting for differences in instance size between stage and prod.

### To address #1 

* By using variables to control desired capacity, we allow ourselves to set variables on the command line to override the values in the terraform scripts.  

* I've separated out "appgroup" scale (1 stage, 3 prod for minimums) from "controllergroup" (*aka, nodes we'll only have one of, such as admin*) scale (1 stage, 1 prod).

* For beginning infrastructure I have desired-capacity mirroring min-size.  

### To address #2 

* I've set variables to stand in for many of our nodes instance size values, wherever the instance size is specced differently in stage vs prod.  (Two sets of variables: processorgroup_instance_size and appgroup_instance_size)

* This will allow us to scale up vertically when needed by simply doing a *./wrapper.sh "apply -var appgroup_instance_size.prod=m3.xlarge" prod webapp* (as an example)

* This will spin up larger infrastructure we've specced out for prod without spending all of Uncle Lonnen's cash on stage.


jdotpz:
Well, our tests are failing, but here are the plans for each role.

https://www.dropbox.com/home/public/terraform-plan-logs-1430831837050517S

phrawzty:
This is failing because you're passing `FAKE` to a `lookup()` function, but since that key doesn't exist, the lookup fails.  The solution is to use an extant key (such as `stage`) so that the lookup can procure a value (or to add FAKE as a key, which would work, but is meh imho).

PR #91, opened 2015-05-05
bug 1152621 - activate nginx configs when nodes are provisioned by ASG

rhelmer:
r? @jdotpz / @phrawzty


puppet/modules/socorro/files/etc_nginx/nginx.conf
```diff
@@ -0,0 +1,44 @@
+# For more information on configuration, see:
```
rhelmer: Note that this file is included because we need to remove the `default_server` that CentOS nginx package ships with, otherwise we can't determine the default server.

We will probably want to tweak things in here such as logging anyway.

jdotpz:
R+ for the moment.

PR #94, opened 2015-05-06
Upsize Leeroy so he doesn't get everyone killed right away

jdotpz:
Upsize to a C3.Large, based from AMI I just snapshotted.  I put it to a 100gb root volume with general SSD, and added an instance store volume.
When it launches I'll add some Swap with the instance store volume.


rhelmer:
r+ not right away

PR #97, opened 2015-05-06
Allow ingress to my coworkers and github

jdotpz:


rhelmer:
r+ letting the riff-raff in

PR #99, opened 2015-05-06
lint wrapper.sh

phrawzty:
Ran the script through a [shell linter](http://shellcheck.net) and cleaned up the warnings (they were all double-quote issues).

@rhelmer @jdotpz `r?`


rhelmer:
r+, let's get that running on travis-ci too!

PR #101, opened 2015-05-08
run shell script linter shellcheck

rhelmer:
r? @jdotpz / @phrawzty

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,11 @@
-sudo: no
+sudo: yes
 language: python
 
-install: false
+install: true
+
+before_script:
+  - wget "https://launchpad.net/ubuntu/+source/shellcheck/0.3.3-1~ubuntu14.04.1/+build/6293671/+files/shellcheck_0.3.3-1%7Eubuntu14.04.1_amd64.deb"
+  - sudo dpkg -i shellcheck_0.3.3-1~ubuntu14.04.1_amd64.deb
 
 script:
   - ./bin/test.sh
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -4,10 +4,15 @@ set -e
 
 TFORM_VERSION="0.4.2_linux_amd64"
 
+# lint check for shell scripts
+shellcheck $(find . -name "*.sh")
+
+# lint check for puppet
 gem install puppet puppet-lint
 puppet-lint --with-filename --no-80chars-check --no-autoloader_layout-check --fail-on-warnings puppet/
 puppet parser validate `find puppet/ -name '*.pp'`
 
+# ensure "terraform plan" returns no errors
 pushd terraform
 wget "https://dl.bintray.com/mitchellh/terraform/terraform_${TFORM_VERSION}.zip"
 unzip -u terraform_${TFORM_VERSION}.zip


.travis.yml
```diff
@@ -1,7 +1,11 @@
-sudo: no
+sudo: yes
 language: python
 
-install: false
+install: true
+
+before_script:
+  - wget "https://launchpad.net/ubuntu/+source/shellcheck/0.3.3-1~ubuntu14.04.1/+build/6293671/+files/shellcheck_0.3.3-1%7Eubuntu14.04.1_amd64.deb"
+  - sudo dpkg -i shellcheck_0.3.3-1~ubuntu14.04.1_amd64.deb
```
rhelmer: This package isn't in stable but there's a PPA. We could install the PPA the usual way and `apt-get install` if we don't want to pin the specific version (apt provides a pinning mechanism too)

rhelmer: We probably should just install the PPA the usual way.

.travis.yml
```diff
@@ -1,7 +1,11 @@
-sudo: no
+sudo: yes
```
rhelmer: I'd rather not turn this on, but the alternative is we either need to compile shellsheck (it is written in Haskell) or check in a binary.

phrawzty:
There's a way to do this that lets us keep `sudo: no` - it's a bit dirty, but if we're already comfortable with installing .deb's from Launchpad as part of our test flow, then we're not far off.  Since this package is just chrome around a binary, we can reasonably obtain the .deb, extract it, and do what we like with that binary without actually installing it to the system.

In Bash form:
```bash
#!/usr/bin/env bash

SC_DEB="shellcheck_0.3.3-1~ubuntu14.04.1_amd64.deb"
SC_URL="https://launchpad.net/ubuntu/+source/shellcheck/0.3.3-1~ubuntu14.04.1/+build/6293671/+files/${SC_DEB}"
SC_DIR="${HOME}/.shellcheck"

mkdir -p "$SC_DIR"
pushd "$SC_DIR"

wget "$SC_URL"
ar x "$SC_DEB"
tar -xf data.tar.xz

export PATH="${PATH}:${SC_DIR}/usr/bin"

shellcheck --version

popd
```

Untested YAML for Travis:
```yaml
sudo: no
install: false
before_script:
  - wget "https://launchpad.net/ubuntu/+source/shellcheck/0.3.3-1~ubuntu14.04.1/+build/6293671/+files/shellcheck_0.3.3-1%7Eubuntu14.04.1_amd64.deb"
  - ar x shellcheck_0.3.3-1~ubuntu14.04.1_amd64.deb
  - tar -xf data.tar.xz
```

And then in `bin/test.sh` alter the shellcheck invocation to point to `$HOME/usr/bin` (or alter `$PATH`).

rhelmer:
@phrawzty that's not a bad idea, I hadn't considered just extracting the binary from the deb.

What advantages are we getting from `sudo: no` right now? I know you can share the checkout dir between builds but we don't need/want that, right?

phrawzty:
> What advantages are we getting from sudo: no right now? I know you can share the checkout dir between builds but we don't need/want that, right?

IIRC builds with `sudo: no` are handled differently by Travis and are much faster overall.

In any case, as you can see from the [Travis run](https://travis-ci.org/mozilla/socorro-infra/builds/62300094), Shellcheck is going to come up with a lot of complaints. I recommend that we examine these complaints, determine their validity, and address them as necessary before landing Shellcheck - otherwise it's going to be a long while before anything *else* passes...

rhelmer:
> In any case, as you can see from the Travis run, Shellcheck is going to come up with a lot of complaints. I recommend that we examine these complaints, determine their validity, and address them as necessary before landing Shellcheck - otherwise it's going to be a long while before anything else passes...

+1

Also note that Shellcheck has a `-e` that allows you to exclude codes from the result (from the manpage: `Explicitly exclude the specified codes from the report. Subsequent -e options are cumulative, but all the codes can be specified at once, comma-separated as a single argument.`)

> IIRC builds with sudo: no are handled differently by Travis and are much faster overall.

OK I'll change this to use the binary, extracting from the deb package would work for now.

rhelmer:
One very interesting thing about Shellcheck is that they have wiki pages with rationale on why the warnings are there, and suggestions for fixing the code.

For instance, the `./bin/test.sh` script that we use for Jenkins has these problems:
```
$ shellcheck bin/test.sh 

In bin/test.sh line 8:
shellcheck $(find . -name "*.sh")
           ^-- SC2046: Quote this to prevent word splitting.


In bin/test.sh line 13:
puppet parser validate `find puppet/ -name '*.pp'`
                       ^-- SC2046: Quote this to prevent word splitting.
                       ^-- SC2006: Use $(..) instead of legacy `..`.


In bin/test.sh line 22:
for role in $(find ./terraform/* -maxdepth 1 -type d); do
            ^-- SC2044: For loops over find output are fragile. Use find -exec or a while read loop.


In bin/test.sh line 23:
    pushd $role
          ^-- SC2086: Double quote to prevent globbing and word splitting.
```

The rationale and example "correct" code is:
https://github.com/koalaman/shellcheck/wiki/Sc2046
https://github.com/koalaman/shellcheck/wiki/SC2006
https://github.com/koalaman/shellcheck/wiki/SC2044
https://github.com/koalaman/shellcheck/wiki/SC2086

An example of applying recommended changes so that shellcheck passes:
```
diff --git a/bin/test.sh b/bin/test.sh
index f48211a..a83bece 100755
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -5,12 +5,14 @@ set -e
 TFORM_VERSION="0.4.2_linux_amd64"
 
 # lint check for shell scripts
-shellcheck $(find . -name "*.sh")
+find . -name "*.sh" | while IFS='' read -r shellscript; do
+  shellcheck "$shellscript"
+done
 
 # lint check for puppet
 gem install puppet puppet-lint
 puppet-lint --with-filename --no-80chars-check --no-autoloader_layout-check --fail-on-warnings puppet/
-puppet parser validate `find puppet/ -name '*.pp'`
+puppet parser validate "$(find puppet/ -name '*.pp')"
 
 # ensure "terraform plan" returns no errors
 pushd terraform
@@ -19,8 +21,8 @@ unzip -u terraform_${TFORM_VERSION}.zip
 ./wrapper.sh symlinks
 popd
 
-for role in $(find ./terraform/* -maxdepth 1 -type d); do
-    pushd $role
+find ./terraform/* -maxdepth 1 -type d | while IFS='' read -r role; do
+    pushd "$role"
     ../terraform plan -var="environment=FAKE" \
                    -var="secret_key=FAKE" \
                    -var="access_key=FAKE" \
```

lonnen:
still desired, but not _that_ desired

phrawzty:
screen shot 2016-10-18 at 09 39 10

PR #102, opened 2015-05-09
fix socket paths to be like socorro RPM

rhelmer:
r? @jdotpz


rhelmer:
Going to carry over r? from the RPM, get this running now.

PR #106, opened 2015-05-12
Envbanner

jdotpz:


phrawzty:
Looks like you've got trailing commits in this PR too...

jdotpz:
But strangely no edits.  I think master already has those landed, and it didn't give me an option to squash them all.

AdrianGaudebert:
@jdotpz that happened to me in the past, to solve it I would create a new branch from master and cherry-pick my commit there.

jdotpz:
Good call, doing that here: https://github.com/mozilla/socorro-infra/pull/110

PR #108, opened 2015-05-12
Create a backup script for jenkins while will swap in new AMIs to the…

jdotpz:
… buildbox launch config

### DIFF
--- a/bin/jenkins/backup_jenkins.sh
+++ b/bin/jenkins/backup_jenkins.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+# Script: backup_buildbox.sh
+# Purpose: Get image of buildbox, apply it to the AS group
+
+############################
+# COMMON FUNCTIONS AND SETTINGS
+# Show usage instructions for noobs
+function show_usage() {
+    echo " =================== ";echo " ";
+    echo " Syntax "
+    echo "USAGE: $0"
+    echo " "
+    exit 0
+}
+
+function error_check() {
+    if [ ${RETURNCODE} -ne 0 ];then
+        echo "`date` -- Error encountered during ${PROGSTEP} : ${RETURNCODE}"
+        if [ "${NOTFATAL}" = "true" ];then
+            echo "Not fatal, continuing"
+            NOTFATAL="false"
+        else
+            echo "Fatal, exiting"
+            echo "===================="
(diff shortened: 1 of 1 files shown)


bin/jenkins/backup_jenkins.sh
```diff
@@ -0,0 +1,86 @@
...
+
+function format_logs() {
+    echo " ";echo " ";echo " ";echo "=================================================";echo " ";echo " ";echo " "
+}
+
+####################################
+## PROGRAM FUNCTIONS
+function create_ami()  {
+    PROGSTEP="Creating AMI"
+    AMINAME="`date +%Y%m%d%H%M`-stage-socorrobuildbox"
```
phrawzty: Maybe put the datestamp at the end of the string instead? I'm not sure if we'd rather sort by date or by image name (type). Thoughts?

bin/jenkins/backup_jenkins.sh
```diff
@@ -0,0 +1,86 @@
...
+# Script: backup_buildbox.sh
+# Purpose: Get image of buildbox, apply it to the AS group
+
+############################
+# COMMON FUNCTIONS AND SETTINGS
+# Show usage instructions for noobs
+function show_usage() {
+        echo " =================== ";echo " ";
+        echo " Syntax "
+        echo " ./deploy-socorro.sh $env (stage or prod)"
```
phrawzty: This will interpolate `$env` which is an empty variable at this point. Prefer:
```bash
echo "USAGE: $0 <stage|prod>"
```
Alternatively, if we we want to future-proof, use `<ENV>` instead of stating environments explicitly.

bin/jenkins/backup_jenkins.sh
jdotpz: Yeah, that ties in with our standard better, good call.

rhelmer:
r+

bin/jenkins/backup_jenkins.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+# Script: backup_buildbox.sh
```
phrawzty: Here you've got `backup_buildbox.sh`, but the script is checked in as `backup_jenkins.sh`; suggest renaming the script to the former as it falls in line with the rest of our nomenclature.

PR #113, opened 2015-05-13
Add a way to manually re-trigger deploys with GITHASH

jdotpz:
When a job fails on the terraform step, it's shameful to have to start all over again and do AMI build and version build.  Further, it's a PITA to manually copy in the payload to Jenkins.

Thus, this makes us able to setup jobs in Jenkins, which you can manually run, supply an github commit hash, and it will then replace the environments AMI's and swap out nodes.

### DIFF
--- a/bin/deploy-socorro.sh
+++ b/bin/deploy-socorro.sh
@@ -11,8 +11,14 @@ if echo $2 | grep "skiprpm" > /dev/null;then
     SKIPRPM="true"
 fi
 if echo $3 | grep "skipami" > /dev/null;then
-    NEWAMI="ami-19102029"
+    NEWAMI="ami-19102029" #just random
     SKIPAMI="true"
+fi
+if [ "$2" = "manual" ];then
+    MANUALPUSH="true" # We're just doing a deploy by HASH
+    HASHTODEPLOY=${GITHUBHASH} # This will be build with parameters for git commit has to build in jenkins
+else
+    MANUALPUSH="false"
 fi
 GITPAYLOAD=${payload}
 RANDOM_STRING="$RANDOM-$RANDOM"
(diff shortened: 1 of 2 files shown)


bin/lib/create_rpm.sh
```diff
@@ -25,7 +25,7 @@ function create_rpm() {
     # Find the rpm file we created
     NEWRPM=`ls -lart /data/socorro/socorro*.rpm|awk '{print $9}'`
     # Get the version of that socorro rpm so we log it nicely.
-    NEWSOCORROVERSION=`ls /home/centos/org.mozilla.crash-stats.packages-public/x86_64/socorro-*.rpm|tail -n1|sed 's/\// /g'|sed 's/\./ /g'|awk '{print $8}'`
+    NEWSOCORROVERSION=`ls -lart /home/centos/org.mozilla.crash-stats.packages-public/x86_64/socorro-*.rpm|tail -n1|sed 's/\// /g'|sed 's/\./ /g'|awk '{print $8}'`
```
rhelmer: why do you need `sed 's/\// /g'|sed 's/\./ /g'` here, just replacing `/` and `.` with a space?

I think `tr` is more readable for this: `tr ['.','/'] ' '`

rhelmer:
r+, take or leave my `tr` vs `sed` comment.

bin/lib/create_rpm.sh
jdotpz: I'll actually clean that up everywhere it exists in the scripts, but push onward with merging this.

bin/deploy-socorro.sh
```diff
@@ -29,7 +35,7 @@ RANDOM_STRING="$RANDOM-$RANDOM"
 function show_usage() {
         echo " =================== ";echo " ";
         echo " Syntax "
-        echo " ./deploy-socorro.sh $env (stage or prod)"
+        echo " ./deploy-socorro.sh $env [manual](optional) [$githash-to-deploy](optional)"
```
phrawzty: Please remove the `(optional)` strings - in *nix, one uses chevrons to denote `<required args>` and brackets to denote `[optional ones]`.

Also, as per my previous recommendation in a different PR, any variable you put in double quotes will be interpolated, which means that Bash will attempt display the values both `$env` and `$githash-to-deploy`, which is almost certainly not the intended outcome here. Your options: escape the `$` signs, use single-quotes for this string, or don't use `$` at all.

Recommend:
```bash
echo "USAGE: $0 <env> [manual] [githash-to-deploy]"
echo "Where <env> is the environment, [manual] is <true|false>, and [githash-to-deploy] is a <git short hash>."
```

PR #114, opened 2015-05-14
Take advantage of a few new wins in terraform v0.5.0

jdotpz:
* Allow terraform to retry 5 times after transient AWS api errors
* Enable 30 second timeout connection draining on our elbs

### DIFF
--- a/terraform/analysis/main.tf
+++ b/terraform/analysis/main.tf
@@ -77,6 +77,8 @@ resource "aws_elb" "elb-socorroanalysis" {
         project = "socorro"
     }
     cross_zone_load_balancing = true
+    connection_draining = true
+    connection_draining_timeout = 30
 }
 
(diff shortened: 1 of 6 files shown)


rhelmer:
@jdotpz so I am confused about the connection draining - does that come into play when making a new ELB, or is it for if terraform changes the AS associated with the ELB?

rhelmer:
+1 on the retries, that's great

jdotpz:
@rhelmer It is setup just once on a new ELB, but prior it was one of those "Do it yourself things" which I kept doing for testing.  Then, when you do an instance deregister from the elb, it doesn't just rip it out, it waits the timeout for connections to drain while it's taking no more new connects.

rhelmer:
r+

phrawzty:
very nice

PR #115, opened 2015-05-15
this lives on jenkins now, no way to keep the git version up to date

rhelmer:
r? @phrawzty @jdotpz


phrawzty:
You'll need to *FAKE* the var for Travis, otherwise quite simple so `r+`.

rhelmer:
@phrawzty hm, we can't fake this like the others since a lookup is being done on it:

```
* lookup: lookup in 'FAKE' failed to find 'us-west-2' in:

${lookup(var.base_ami, var.region)}
```

I'm not sure how to override this from the command line, in `terraform.tfvars` we do `base_ami.us-west-2="..."` but if I try to override it like `-var="base_ami.us-west-2=FAKE"` then I get:

```
  * Required variable not set: base_ami
```

jdotpz:
```./wrapper.sh "apply -var base_ami.us-west-2=FAKE" stage admin```

That will work to override.
(Output from a plan)
-/+ aws_launch_configuration.lc-socorroadmin
    associate_public_ip_address: "true" => "1"
    ebs_block_device.#:          "0" => "<computed>"
    ebs_optimized:               "false" => "<computed>"
    iam_instance_profile:        "generic" => "generic"
    image_id:                    "ami-2da1931d" => "FAKE" (forces new resource)

**Comment by phrawzty** [comment]
I think we're over-thinking here. :smile: The only reason `$base_ami` was ever a `lookup()` in the first place was because the intent was to support multiple regions for this value; since this is no longer the case, we should just make `$base_ami` a basic variable, and replace the `lookup()` with a reference to the same.

File: bin/test.sh [hunk 1]
```diff
@@ -25,6 +25,7 @@ for role in $(find ./terraform/* -maxdepth 1 -type d); do
                    -var="analysis_cert=FAKE" \
                    -var="buildbox_cert=FAKE" \
                    -var="webapp_cert=FAKE" \
-                   -var="rds_root_password=FAKE"
+                   -var="rds_root_password=FAKE" \
+                   -var="base_ami.us-west-2=FAKE"
```
phrawzty: `.us-west-2` can be removed now.

phrawzty:
We let this PR slide but IMHO it's a step in the right direction.

Note that for infra which isn't managed by Buildbox (Elasticsearch, for example) if updates are made to Puppet, one must manually build an AMI and then pass it as a var to Terraform explicitly. Might there be a way to store the *most recent* AMI IDs (i.e. both Base and Buildbox) somewhere so that it can be scripted? ping @rhelmer @jdotpz

PR #116, opened 2015-05-15
[DO NOT MERGE] add support for different ES roles

phrawzty:
There are ASG and LC resource blocks for each of the three types of ES role (master, interface, and data, respectively). I had to pass this bit of information to userdata somehow, so I exploited the `socorro_role` positional argument to pass additional data (i.e. `'elasticsearch FACTER_elasticsearch_role=master'`). This is gross but it works for now. :unamused: 

As a side-effect, I appear to have proposed a model for splitting out values based on environment, as seen in `variables.tf`. It's functional and readable, so that's nice, but I'm not married to it.

Finally, I had to change the Terraform labels to `socorroelasticsearch2` so that the resources wouldn't collide with the existing `socorroelasticsearch` resources; before we merge, the latter resources will need to be removed entirely, then these can be renamed as appropriate.

Anyways, this is a WIP and hasn't been tested even once, so please ignore for now. :grinning:

### DIFF
--- a/puppet/modules/socorro/templates/etc_elasticsearch/elasticsearch.yml.erb
+++ b/puppet/modules/socorro/templates/etc_elasticsearch/elasticsearch.yml.erb
@@ -1,0 +1,25 @@
+---
+name: "<%= @hostname %>"
+cluster:
+    name: "sss_<%= @environment %>"
+discovery:
+    zen:
+        minimum_master_hosts: 2
+    type: "ec2"
+    ec2:
+        tag:
+            Environment: "<%= @environment %>"
+            role: "elasticsearch"
+index:
+    number_of_replicas: 1
+action:
+    disable_shutdown: true
+    disable_delete_all_indices: true
+cloud:
+    aws:
+        region: "<%= @ec2_placement_availability_zone.match(/\w+\-\w+/) %>"
+node:
+    master: <%= @es_master %>
+    data: <%= @es_data %>
+http:
+    enabled: <%= @es_interface %>
--- a/terraform/elasticsearch/main.tf
+++ b/terraform/elasticsearch/main.tf
@@ -4,8 +4,8 @@ provider "aws" {
     secret_key = "${var.secret_key}"
 }
 
-resource "aws_security_group" "elb-socorroelasticsearch-sg" {
-    name = "elb-socorroelasticsearch-${var.environment}-sg"
+resource "aws_security_group" "elb-socorroelasticsearch2-sg" {
+    name = "elb-socorroelasticsearch2-${var.environment}-sg"
     description = "Allow internal access to Elasticsearch."
     ingress {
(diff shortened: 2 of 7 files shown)


puppet/modules/socorro/templates/etc_elasticsearch/elasticsearch.yml.erb
```diff
@@ -0,0 +1,25 @@
+---
+name: "<%= @hostname %>"
+cluster:
+    name: "sss_<%= @environment %>"
+discovery:
+    zen:
+        minimum_master_hosts: 2
+    type: "ec2"
+    ec2:
+        tag:
```
phrawzty: The AWS plugin can read tags! This helps to ensure that we don't join the clusters from different environments together.

puppet/modules/socorro/templates/etc_elasticsearch/elasticsearch.yml.erb
```diff
@@ -0,0 +1,25 @@
...
+            Environment: "<%= @environment %>"
+            role: "elasticsearch"
+index:
+    number_of_replicas: 1
+action:
+    disable_shutdown: true
+    disable_delete_all_indices: true
+cloud:
+    aws:
+        region: "<%= @ec2_placement_availability_zone.match(/\w+\-\w+/) %>"
```
phrawzty: `ec2_placement_availability_zone` is a fact with a value in the form `us-west-2`; the AWS plugin only wants the region, which is all this will `match()`.

terraform/variables.tf
```diff
@@ -36,6 +36,42 @@ variable "elb_master_web_sg_id" {
...
+variable "es_data_ec2_type" {
+    default = {
+        stage = "i2.xlarge"
+        prod = "i2.2xlarge"
+    }
+}
+variable "es_data_num" {
+    default = {
+        stage = "3"
+        prod = "9"
```
phrawzty: These values are based on having enough disk space to support the environments as they exist today and may not be valid once we trim the raw crashes out.

terraform/elasticsearch/main.tf
```diff
@@ -4,8 +4,8 @@ provider "aws" {
     secret_key = "${var.secret_key}"
 }
 
-resource "aws_security_group" "elb-socorroelasticsearch-sg" {
-    name = "elb-socorroelasticsearch-${var.environment}-sg"
+resource "aws_security_group" "elb-socorroelasticsearch2-sg" {
+    name = "elb-socorroelasticsearch2-${var.environment}-sg"
```
jdotpz: Should be ${var.environment}-socorroelasticsearch2

phrawzty:
Upstream has drifted too far for this PR to be easily salvaged. Closing and opening a new PR based off of d9362ef.

PR #119, opened 2015-05-18
[DO NOT MERGE] check in aws migration script

rhelmer:
r? @phrawzty @jdotpz - this isn't quite ready yet, and it's a bit quick&dirty, but I wanted to make sure there's some public record of the AWS RDS migration work.

This passes shellcheck - note the use of the per-line disable directive.


archive/aws_migration.sh
```diff
@@ -0,0 +1,110 @@
...
+    TABLES=$1
+    if [ ! -f "$TABLES" ]; then
+        echo "Error: $TABLES must exist"
+        return 1
+    fi
+    time pg_dump \
+      --format=c \
+      --no-owner \
+      --no-privileges \
+      # shellcheck disable=SC2059 - we actually want word splitting here
```
phrawzty: :+1:

phrawzty:
Is the line order of the `.txt` files significant? If so that should probably be noted in some way.

phrawzty:
Out of curiousity, what happens if a step fails and the script bails on `set -e`? Is it safe / possible to run again?

rhelmer:
> Is the line order of the .txt files significant? If so that should probably be noted in some way.

Yes, I'll note this.

> Out of curiousity, what happens if a step fails and the script bails on set -e? Is it safe / possible to run again?

It needs to be recovered manually - most likely by starting over (drop database and drop user from RDS DB, run script again)

phrawzty:
> It needs to be recovered manually - most likely by starting over (drop database and drop user from RDS DB, run script again)

Would be it possible to add some output here to help guide the user? Like "failed <here> so investigate as of that point" type thing?

lonnen:
r+, merge when ready

PR #120, opened 2015-05-18
Create a script to automagically cleanup AWS

jdotpz:
* Check for unused, 'available' volumes, and delete them.
* Check for unused launch configs and delete them.

### DIFF
--- a/bin/jenkins/aws_cleanup.sh
+++ b/bin/jenkins/aws_cleanup.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+# Script: aws_cleanup.sh
+# Purpose: Cleanup unused volumes, launch configs, etc
+
+############################
+# COMMON FUNCTIONS AND SETTINGS
+# Show usage instructions for noobs
+function show_usage() {
+    echo " =================== ";echo " ";
+    echo " Syntax "
+    echo "USAGE: $0"
+    echo " "
+    exit 0
+}
+
+function error_check() {
+    if [ ${RETURNCODE} -ne 0 ];then
+        echo "`date` -- Error encountered during ${PROGSTEP} : ${RETURNCODE}"
+        if [ "${NOTFATAL}" = "true" ];then
+            echo "Not fatal, continuing"
+            NOTFATAL="false"
+        else
+            echo "Fatal, exiting"
+            echo "===================="
+            echo "Instances which may need to be terminated manually: ${INITIALINSTANCES}"
+            exit 1
+        fi
+    fi
+}
+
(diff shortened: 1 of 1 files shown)


bin/jenkins/aws_cleanup.sh
```diff
@@ -0,0 +1,69 @@
...
+    echo " ";echo " ";echo " ";echo "=================================================";echo " ";echo " ";echo " "
+    aws ec2 describe-volumes --filters Name=status,Values="available"
+}
+
+############################
+# PROGRAM FUNCTIONS
+function cleanup_unused_volumes() {
+    PROGSTEP="Cleaning up unused volumes"
+    echo "`date` -- Getting list of unused volumes"
+    UNUSEDVOLUMES=`aws ec2 describe-volumes --filters Name=status,Values="available"|grep VolumeId|sed 's/"/ /g'|awk '{print $3}'`
```
phrawzty: This will certainly create a list of all the "available" volumes in a given zone; however, what about instances where a volume is currently unused but is being retained for future use? (There's one such example sitting in us-west-2 right now, in fact.)

phrawzty:
I whole-heartedly support a method for keeping our AWS infra clean of cruft, but I'm concerned about the possible repercussions of removing things based solely on the criteria that they happen to be unused at the moment this script happens to be run. Imho we'll need to reflect on some more conditions for removal / preservation.  For example:
* Younger than X? Keep.
* Older than X? Remove.
* Has tag "NODELETE: true"? Keep even if Older than X.
* etc.

Thoughts?

phrawzty:
Shellcheck throws a bunch of warnings on this, mostly related to using backticks instead of `$()` and double-quoting variables to avoid splits - you may wish to take a look at that, too.

jdotpz:
Hmmm, yeah, if we're going to store volumes we need, then this would be a little too reaper-riffic.
The tag NODELETE is a darn good idea.  I'd be ok with adding that tag to anything we don't want removed, and have that be the canonical "check if i can delete this" flag.

phrawzty:
> I'd be ok with adding that tag to anything we don't want removed, and have that be the canonical "check if i can delete this" flag.

This is fine as long as everybody is on the same page - and that includes people who aren't on our team, which makes it tricky.  A safer option might be to have an opt-in tag (instead of opt-out), such that we can reap our properly tagged stuff automatically while leaving other people's stuff alone.

In any case, we need to determine if and how to apply tags to resources which Terraform does not manage directly - recall that the Volumes (specifically) are created via ASG, not TF...

PR #122, opened 2015-05-19
Use instanceid as hostname

jdotpz:


puppet/modules/socorro/manifests/role/common.pp
```diff
@@ -8,4 +8,25 @@
...
   }
 
+  exec {
+    'install-ec2-metadata':
+      path    => '/usr/bin',
+      command => '/bin/wget http://s3.amazonaws.com/ec2metadata/ec2-metadata -P /bin/',
+      notify  => Exec['set-hostname'],
+      creates =>
+  }
+
```
rhelmer: This should happen in the packer phase, otherwise this failing will cause nodes spun up by the AS to fail (and it just fits with our "immutable" style better)

puppet/modules/socorro/manifests/role/common.pp
```diff
@@ -8,4 +8,25 @@
...
+  exec {
+    'install-ec2-metadata':
+      path    => '/usr/bin',
+      command => '/bin/wget http://s3.amazonaws.com/ec2metadata/ec2-metadata -P /bin/',
+      notify  => Exec['set-hostname'],
+      creates => '/bin/ec2-metadata'
+  }
+
```
rhelmer: Is this supposed to be in here twice? :)

puppet/modules/socorro/manifests/role/common.pp
jdotpz: It can't happen in packer phase, because it has no clue what node(s) will be spun up off that image.
It shouldn't fail on node bootstrap.

rhelmer: > It can't happen in packer phase, because it has no clue what node(s) will be spun up off that image.

Oh I meant the `wget` bit, not the actual executing of the thing.

rhelmer:
r+

rhelmer:
Ideally we'd package this (/me invokes @phrawzty) but it's a simple shell script on S3, as distributed by Amazon, so I think we're better off just doing the single `wget` and be done with it.

PR #124, opened 2015-05-19
bug 1146447 - re-join the consul cluster when restarting in server mode

rhelmer:
r? @phrawzty / @jdotpz 

The problem with consul servers right now is that they do `consul join` before restarting in server mode, which causes them to leave the cluster.

They need to do a `consul join` after restarting in server mode, and consul should come up automagically.

### DIFF
--- a/puppet/modules/socorro/manifests/role/consul.pp
+++ b/puppet/modules/socorro/manifests/role/consul.pp
@@ -18,4 +18,12 @@ include socorro::role::common
       mode   => '0640';
   }
 
+  # We expect this to come from the secret S3 bucket
+  $consul_hostname = hiera("${::environment}/consul_hostname")
+  exec {
+    'join_consul_server_cluster':
+      command  => "/usr/bin/consul join ${consul_hostname}",
+      requires => Service['consul'];
+  }
+
 }


puppet/modules/socorro/manifests/role/consul.pp
```diff
@@ -18,4 +18,12 @@
       mode   => '0640';
   }
 
+  # We expect this to come from the secret S3 bucket
+  $consul_hostname = hiera("${::environment}/consul_hostname")
+  exec {
+      'join_consul_cluster':
```
phrawzty: Style nit: check your whitespace and arrow alignment in this block.

phrawzty: Actual problem: This will result in a namespace collision with `Exec['join_consul_cluster']` defined in `socorro::role::common` (included in all roles).

phrawzty:
Both `Service['consul']` and `Exec['join_consul_cluster']` depend on `File['/etc/consul/server.json']`, but there is no explicit relationship between the Service and the Exec, which means the order in which those will be applied isn't deterministic in any meaningful way.

Furthermore, note that `Service['consul']` is ensured and enabled in the base `socorro` class (Packer phase), which means that it's enabled when ASG spins up a node (Provision phase) and thus *prior* to the existence of `server.json`. The subscription was supposed to address this problem by restarting the Service when that configuration file was applied; however, as you've discovered, simply restarting the service apparently isn't enough - we also need to issue a `consul join`.

I therefore propose the following instead:
* Rename the `Exec` so as to avoid the [namespace collision](https://github.com/mozilla/socorro-infra/pull/124#discussion_r30697601).
* Replace the File requirement in the Exec with the Service, effectively creating the chain: File → Service → Exec.

This *should* address this issue in question. :smile:

phrawzty:
`r+` on c1ef005

PR #126, opened 2015-05-20
disable symbol cache

rhelmer:
r? @jdotpz


jdotpz:
r+, merging after travis says...well there travis goes.

PR #127, opened 2015-05-20
Wait longer to scale down

jdotpz:
To help ensure we don't have autoscaling groups scaling down during a push, this changes the low cpu alert in cloudwatch for a given as group to wait for 3 periods of 5 minutes instead of just 5 minutes.


rhelmer:
r+

PR #128, opened 2015-05-20
Install datadog agent on all nodes

jdotpz:
This will install datadog agent on the packer step, and then drop in the config files with api key on launch step.

### DIFF
--- a/puppet/modules/socorro/manifests/init.pp
+++ b/puppet/modules/socorro/manifests/init.pp
@@ -1,6 +1,11 @@
 # Elements common to all Socorro boxes.
 class socorro {
 
+  # Setup run stages to set hostname nice and early
+  stage { 'prep':
+    before => Stage['main'],
+  }
+
   # Yum will sometimes resolve and attempt to use IPv6 addresses. This is a
   # problem in AWS and can cause random errors (fun!).
   $yum_ipv4_exec = $::bios_version ? {
@@ -25,5 +30,12 @@ class socorro {
       ensure  => running,
       enable  => true,
       require => File['sshd_config']
+  }
+
+  exec {
+    'set-hostname':
+      path    => '/usr/bin',
+      command => '/bin/ec2-metadata -i |cut -d " " -f 2 > /etc/hostname && /bin/hostname -F /etc/hostname',
+      stage   => prep
   }
 }
--- a/puppet/modules/socorro/manifests/init.pp
+++ b/puppet/modules/socorro/manifests/init.pp
@@ -1,11 +1,6 @@
 # Elements common to all Socorro boxes.
 class socorro {
 
-  # Setup run stages to set hostname nice and early
-  stage { 'prep':
-    before => Stage['main'],
-  }
-
   # Yum will sometimes resolve and attempt to use IPv6 addresses. This is a
   # problem in AWS and can cause random errors (fun!).
   $yum_ipv4_exec = $::bios_version ? {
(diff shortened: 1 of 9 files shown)


puppet/modules/socorro/manifests/init.pp
```diff
@@ -16,9 +16,8 @@
     'consul':
       ensure  => running,
       enable  => true,
-      require => File[
-        '/etc/consul/common.json',
-        '/etc/sysconfig/consul'
+      require => File['/etc/consul/common.json',
+                      '/etc/sysconfig/consul'
```
phrawzty: Please undo this modification - the syntax was correct before.

puppet/modules/socorro/manifests/packer/base.pp
```diff
@@ -185,4 +185,22 @@
       creates => '/bin/ec2-metadata',
       require => Package['wget']
   }
+
+  yumrepo {'datadog':
```
phrawzty: Please put a space between the `{` and the `'`.

puppet/modules/socorro/manifests/role/admin.pp
```diff
@@ -12,4 +12,14 @@
       require => Exec['join_consul_cluster']
   }
 
+  file {
+    '/etc/dd-agent/datadog.conf':
+      mode    => '0600',
+      owner   => root,
```
phrawzty: Please quote all *non-keyword* strings; recall that non-interpolated strings should be enclosed in single quotes.

puppet/modules/socorro/manifests/role/common.pp
```diff
@@ -24,9 +24,23 @@
...
-    'set-hostname':
-      path    => '/usr/bin',
-      command => '/bin/ec2-metadata -i |cut -d " " -f 2 > /etc/hostname && /bin/hostname -F /etc/hostname'
+  # Datadog agent install
+  $datadog_api_key=hiera("${::environment}/datadog_api_key")
+
+  file {
+    '/etc/dd-agent/datadog.conf':
+      mode    => '0640',
+      owner   => dd-agent,
```
phrawzty: Please quote all non-keyword strings; recall that non-interpolated strings should be enclosed in single quotes.

phrawzty:
You created a file path `puppet/modules/socorro/files/etc_dd_agent/` which translates to `/etc/dd_agent/` on the target system; however, based on [this resource declaration](https://github.com/mozilla/socorro-infra/pull/128/files#diff-2fab317501f7cff6be5479cf54911366R14) and the [datadog docs](http://docs.datadoghq.com/guides/basic_agent_usage/centos/) the actual path is `/etc/dd-agent/`. Please modify your Puppet path to use the dash instead of the underscore accordingly.

(As an aside, I've always disliked this underscore-for-paths naming scheme - why don't we just use proper paths? :no_good:)

puppet/modules/socorro/templates/etc_dd_agent/datadog.conf.erb
```diff
@@ -0,0 +1,204 @@
...
+# log_level: INFO
+
+# collector_log_file: /var/log/datadog/collector.log
+# forwarder_log_file: /var/log/datadog/forwarder.log
+# dogstatsd_log_file: /var/log/datadog/dogstatsd.log
+
+# if syslog is enabled but a host and port are not set, a local domain socket
+# connection will be attempted
+#
+# log_to_syslog: yes
```
phrawzty: Do we need to uncomment this or is `yes` the default?

phrawzty:
The [datadog documentation](http://docs.datadoghq.com/guides/agent_checks/) suggests that the YAML configuration files for individual checks should have the `.yaml` extension. I'm guessing that it's not obligatory (since you didn't use the extension, heh), but we should probably stick with the established best practice unless we've got a reason to diverge.

puppet/modules/socorro/manifests/packer/base.pp
```diff
@@ -185,4 +185,22 @@
...
+    enabled  => 1,
+    gpgcheck => 0,
+    descr    => 'Datadog, Inc.',
+    baseurl  => 'http://yum.datadoghq.com/rpm/x86/'
+  }
+
+  package { 'datadog-agent-base':
+    ensure => absent,
```
phrawzty: Why absent?

puppet/modules/socorro/manifests/packer/base.pp
```diff
@@ -185,4 +185,22 @@
       creates => '/bin/ec2-metadata',
       require => Package['wget']
   }
+
+  yumrepo {'datadog':
+    enabled  => 1,
+    gpgcheck => 0,
```
phrawzty: Installing unsigned RPMs over cleartext HTTP.  Awesome.

I realise this isn't your choice - shame on Datadog. :disappointed:

PR #130, opened 2015-05-20
Generate a before and after report of infrastructure

jdotpz:
As exemplified at the end of this Jenkins run ( https://leeroy.mocotoolsstaging.net/job/socorro-stage-replace-ami/17/console ) , this does a Before/After report on nodes, launch configs, and elbs.


phrawzty:
:+1: Having a simple before-and-after view could definitely come in handy if something really unexpected happens.

`r+` on 903fb48.

PR #131, opened 2015-05-20
nobug; set hostname to ec2_instance_id, etc.

phrawzty:
This addresses some concerns with #122:
* The Instance ID is a fact (`ec2_instance_id`) so we don't need a script to obtain it.
* The hostname needs to be set *before* we start sending logs upstream.
* The `ec2-metadata` script is now obtained over HTTPS instead of cleartext.
* Introduced a more canonical usage of Puppet in general.

I tagged this as DO NOT MERGE for now because I'm not sure if we want to bother with the `ec2-metadata` script at all (it appears to only be used in the context of setting the hostname).  I'll remove the related resources if we don't actually need the script.

`r?` @rhelmer @jdotpz

### DIFF
--- a/puppet/modules/socorro/manifests/packer/base.pp
+++ b/puppet/modules/socorro/manifests/packer/base.pp
@@ -178,11 +178,22 @@ class socorro::packer::base {
       mode   => '0644'
   }
 
+  # This little script is used to quickly pull out EC2 metadata via CLI.
+  # TODO: Package it properly.
   exec {
     'install-ec2-metadata':
-      path    => '/usr/bin',
-      command => '/bin/wget http://s3.amazonaws.com/ec2metadata/ec2-metadata -P /bin/ && /usr/bin/chmod 755 /bin/ec2-metadata',
-      creates => '/bin/ec2-metadata',
-      require => Package['wget']
+      path    => '/bin',
+      cwd     => '/bin',
+      command => 'curl -O https://s3.amazonaws.com/ec2metadata/ec2-metadata',
+      creates => '/bin/ec2-metadata'
   }
+
+  file {
+    '/bin/ec2-metadata':
+      owner   => 'root',
+      group   => 'root',
+      mode    => '0755',
+      require => Exec['install-ec2-metadata']
+  }
+
 }
--- a/puppet/modules/socorro/manifests/role/common.pp
+++ b/puppet/modules/socorro/manifests/role/common.pp
@@ -1,11 +1,27 @@
 # Elements common to all Socorro roles.
 class socorro::role::common {
 
+  # Ensure that the hostname is the same as the EC2 instance ID.
+  file {
+    '/etc/hostname':
+      mode    => '0644',
+      owner   => 'root',
+      group   => 'root',
+      content => $::ec2_instance_id
+  }
+
+  exec {
+    'set-hostname':
+      path    => '/bin:/usr/bin',
+      command => 'hostname -F /etc/hostname',
+      require => File['/etc/hostname']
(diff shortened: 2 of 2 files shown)


jdotpz:
One vote in favor of keeping tags is that Datadog will query that ec2-metadata locally to determine the instances tags.

puppet/modules/socorro/manifests/role/common.pp
```diff
@@ -1,11 +1,27 @@
 # Elements common to all Socorro roles.
 class socorro::role::common {
 
+  # Ensure that the hostname is the same as the EC2 instance ID.
+  file {
+    '/etc/hostname':
+      mode    => '0644',
+      owner   => 'root',
+      group   => 'root',
+      content => $::ec2_instance_id
```
jdotpz: +1

puppet/modules/socorro/manifests/role/common.pp
```diff
@@ -1,11 +1,27 @@
...
+  }
+
+  exec {
+    'set-hostname':
+      path    => '/bin:/usr/bin',
+      command => 'hostname -F /etc/hostname',
+      require => File['/etc/hostname']
```
jdotpz: I'm wondering if we should define a prep phase and put this set-hostname in that prep phase.

jdotpz: I'm playing with that approach here: https://github.com/mozilla/socorro-infra/pull/128/files#diff-e2147d10f319bfbfcf35f26b6e7b8684R4

jdotpz: Meh, I don't want to have to start using firm classes to use run stages.  I think the other option is we can put this in the userdata script.

phrawzty: Yeah I specifically avoided implementing stages for this because, well, it's a Pandora's box that I'm not sure we want to open. :spaghetti:

jdotpz:
We can do it here, or we can probably do it in the cloud init script.

phrawzty:
> We can do it here, or we can probably do it in the cloud init script.

I'd prefer that we merge this PR for now (at least) since it puts us in a more sustainable position given the status quo; we can make a more dramatic change (i.e. doing this before Puppet in cloud-init) at a later date.

jdotpz:
+1, better is better.

puppet/modules/socorro/manifests/packer/base.pp
```diff
@@ -178,11 +178,22 @@
...
   }
 
+  # This little script is used to quickly pull out EC2 metadata via CLI.
+  # TODO: Package it properly.
   exec {
     'install-ec2-metadata':
-      path    => '/usr/bin',
-      command => '/bin/wget http://s3.amazonaws.com/ec2metadata/ec2-metadata -P /bin/ && /usr/bin/chmod 755 /bin/ec2-metadata',
-      creates => '/bin/ec2-metadata',
-      require => Package['wget']
```
jdotpz: , at the end of 186.

jdotpz:
Outside of the missing comma, +1's.

puppet/modules/socorro/manifests/packer/base.pp
phrawzty: Er, this line was removed, so, I can't add a comma to it.

rhelmer: I said add a comma

phrawzty: :trollface:

PR #132, opened 2015-05-21
extend ES support; roles, etc

phrawzty:
There are ASG and LC resource blocks for each of the three types of ES role (master, interface, and data, respectively). I had to pass this bit of information to userdata somehow, so I exploited the `socorro_role` positional argument to pass additional data (i.e. `'elasticsearch FACTER_elasticsearch_role=master'`). This is gross but it works for now. :unamused: 

As a side-effect, I appear to have proposed a model for splitting out values based on environment, as seen in `variables.tf`. It's functional and readable, so that's nice, but I'm not married to it.

Finally, I had to change the Terraform labels to `socorroelasticsearch2` so that the resources wouldn't collide with the existing `socorroelasticsearch` resources; before we merge, the latter resources will need to be removed entirely and the former will be renamed as appropriate.

I tested this against AMI ID `ami-2fe2df1f` (built from this branch) and the nodes came up, ES started, discovery occurred, and the cluster came up :green_apple: as expected.

It's also worth noting that this PR introduces another IAM role specific to Elasticsearch (for enabling auto-discovery). As we are not currently managing IAM via Terraform, the details of that role are not present in this PR - a conversation for another time, perhaps. :wink: 

This PR also satisfies [bug 1167134](https://bugzilla.mozilla.org/show_bug.cgi?id=1167134) by installing OpenJDK 1.8 in the Base AMI.

`r?` @rhelmer @jdotpz

### DIFF
--- a/puppet/modules/socorro/manifests/packer/base.pp
+++ b/puppet/modules/socorro/manifests/packer/base.pp
@@ -43,7 +43,7 @@ class socorro::packer::base {
     [
       'bind-utils',
       'consul-ui',
-      'java-1.7.0-openjdk',
+      'java-1.8.0-openjdk',
       'mod_wsgi',
       'nginx',
       'php-cli',
@@ -75,8 +75,14 @@ class socorro::packer::base {
       ensure  => latest,
       require => [
         Yumrepo['elasticsearch'],
-        Package['java-1.7.0-openjdk']
+        Package['java-1.8.0-openjdk']
       ]
+  }
+
+  package {
+    'elasticsearch-plugin-cloud-aws':
+      ensure  => latest,
+      require => Package['elasticsearch']
   }
 
   file {
--- a/puppet/modules/socorro/templates/etc_elasticsearch/elasticsearch.yml.erb
+++ b/puppet/modules/socorro/templates/etc_elasticsearch/elasticsearch.yml.erb
@@ -1,0 +1,25 @@
+---
+name: "<%= @ec2_instance_id %>"
+cluster:
+    name: "sss_<%= @environment %>"
+discovery:
+    zen:
+        minimum_master_hosts: 2
+    type: "ec2"
+    ec2:
+        tag:
+            Environment: "<%= @environment %>"
+            role: "elasticsearch"
+index:
+    number_of_replicas: 1
+action:
+    disable_shutdown: true
+    disable_delete_all_indices: true
+cloud:
+    aws:
+        region: "<%= @ec2_placement_availability_zone.chop %>"
+node:
+    master: <%= @es_master %>
+    data: <%= @es_data %>
+http:
+    enabled: <%= @es_interface %>
--- a/terraform/elasticsearch/main.tf
+++ b/terraform/elasticsearch/main.tf
@@ -4,8 +4,8 @@ provider "aws" {
     secret_key = "${var.secret_key}"
 }
 
-resource "aws_security_group" "elb-socorroelasticsearch2-sg" {
-    name = "elb-socorroelasticsearch2-${var.environment}-sg"
+resource "aws_security_group" "elb-socorroes-sg" {
+    name = "elb-socorroes-${var.environment}-sg"
     description = "Allow ELB access to Elasticsearch."
     ingress {
         from_port = 9200
@@ -33,8 +33,8 @@ resource "aws_security_group" "elb-socorroelasticsearch2-sg" {
     }
 }
 
-resource "aws_security_group" "ec2-socorroelasticsearch2-sg" {
-    name = "ec2-socorroelasticsearch2-${var.environment}-sg"
+resource "aws_security_group" "ec2-socorroes-sg" {
+    name = "ec2-socorroes-${var.environment}-sg"
     description = "Allow internal access to ES, etc."
     ingress {
         from_port = "${var.alt_ssh_port}"
@@ -65,7 +65,7 @@ resource "aws_security_group" "ec2-socorroelasticsearch2-sg" {
         to_port = 9200
         protocol = "tcp"
         security_groups = [
-            "${aws_security_group.elb-socorroelasticsearch2-sg.id}"
+            "${aws_security_group.elb-socorroes-sg.id}"
         ]
     }
     egress {
@@ -94,8 +94,8 @@ resource "aws_security_group" "ec2-socorroelasticsearch2-sg" {
     }
 }
 
-resource "aws_elb" "elb-socorroelasticsearch2" {
-    name = "elb-${var.environment}-socorroelasticsearch2"
+resource "aws_elb" "elb-socorroes" {
+    name = "elb-${var.environment}-socorroes"
     internal = true
     subnets = ["${split(",", var.subnets)}"]
     listener {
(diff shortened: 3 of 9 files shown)


phrawzty:
Travis is failing on 2a61e1e because I haven't modified the test to supply the appropriate FAKE data. I'll push a fix for this soon.

puppet/modules/socorro/manifests/packer/base.pp
```diff
@@ -75,10 +75,16 @@
...
         Yumrepo['elasticsearch'],
-        Package['java-1.7.0-openjdk']
+        Package['java-1.8.0-openjdk']
       ]
   }
 
+  package {
+    'elasticsearch-plugin-cloud-aws':
+      ensure  => latest,
+      require => Package['elasticsearch']
```
jdotpz: Would it make more sense to put this in the role specific to ES?  No other nodes will need it.

phrawzty: We put *everything* in the base package - roles are for service configuration only.

... in fact I'm pretty sure that you're the one who told me that. :smile:

phrawzty:
Concerning the Instance types and counts, those are preliminary and we'll need to do some testing and tuning.  Same goes for the Elasticsearch config itself - the current config is functional but not optimal.

terraform/elasticsearch/main.tf
```diff
@@ -106,24 +106,88 @@ resource "aws_elb" "elb-socorroelasticsearch" {
...
-    instance_type = "t2.micro"
+    instance_type = "${lookup(var.es_master_ec2_type, var.environment)}"
     key_name = "${lookup(var.ssh_key_name, var.region)}"
-    iam_instance_profile = "generic"
+    iam_instance_profile = "socorro_elasticsearch"
     associate_public_ip_address = true
     security_groups = [
-        "${aws_security_group.elb-socorroelasticsearch-sg.id}",
-        "${aws_security_group.ec2-socorroelasticsearch-sg.id}"
+        "${aws_security_group.elb-socorroelasticsearch2-sg.id}",
```
jdotpz: This is going to add EC2 servers to an ELB security group.  Instead, you can add ingress rules to the ec2-sg allowing ports from the elb-sg.

An example of that is : https://github.com/mozilla/socorro-infra/blob/master/terraform/buildbox/main.tf#L60
(Same comment on the other launch configs)

terraform/elasticsearch/main.tf
```diff
@@ -131,15 +195,47 @@ resource "aws_autoscaling_group" "as-socorroelasticsearch" {
...
     depends_on = [
-        "aws_launch_configuration.lc-socorroelasticsearch"
+        "aws_launch_configuration.lc-socorroelasticsearch2-interface"
     ]
-    launch_configuration = "${aws_launch_configuration.lc-socorroelasticsearch.id}"
-    max_size = 10
-    min_size = 1
-    desired_capacity = 1
+    launch_configuration = "${aws_launch_configuration.lc-socorroelasticsearch2-interface.id}"
+    max_size = "${lookup(var.es_interface_num, var.environment)}"
```
jdotpz: +1, love it!

terraform/elasticsearch/main.tf
```diff
@@ -97,7 +97,7 @@ resource "aws_elb" "elb-socorroelasticsearch" {
         lb_protocol = "http"
     }
     security_groups = [
-        "${aws_security_group.elb-socorroelasticsearch-sg.id}"
+        "${aws_security_group.elb-socorroelasticsearch2-sg.id}"
     ]
     tags {
         Environment = "${var.environment}"
```
jdotpz: Not sure if it's really that needed, but you can now turn on connection draining and, iirc, cross zone load balancing for elbs these days.

puppet/modules/socorro/manifests/packer/base.pp
jdotpz: At the time I was talking about on buildbox, but this can't really hurt anything for base, so that works for me. :)

rhelmer: Yeah we don't need Java or ES everywhere either but we are just baking everything into the AMI and only starting services/dropping in files. *Sorta* immutable infra :)

puppet/modules/socorro/templates/etc_elasticsearch/elasticsearch.yml.erb
```diff
@@ -0,0 +1,25 @@
+---
+name: "<%= @ec2_instance_id %>"
+cluster:
+    name: "sss_<%= @environment %>"
```
rhelmer: Why is it named this, still a reference to the original "socorro search service"?

terraform/elasticsearch/main.tf
```diff
@@ -4,8 +4,8 @@ provider "aws" {
     secret_key = "${var.secret_key}"
 }
 
-resource "aws_security_group" "elb-socorroelasticsearch-sg" {
-    name = "elb-socorroelasticsearch-${var.environment}-sg"
+resource "aws_security_group" "elb-socorroelasticsearch2-sg" {
```
rhelmer: Is there more intuitive name we could use instead of `2`?

jdotpz: It's just going to change back to socorroelasticsearch once we merge and replace.  This way, he's not stomping on existing infra when he's testing all of the things.

terraform/elasticsearch/main.tf
```diff
@@ -131,15 +195,47 @@ resource "aws_autoscaling_group" "as-socorroelasticsearch" {
...
-    max_size = 10
-    min_size = 1
-    desired_capacity = 1
+    launch_configuration = "${aws_launch_configuration.lc-socorroelasticsearch2-interface.id}"
+    max_size = "${lookup(var.es_interface_num, var.environment)}"
+    min_size = "${lookup(var.es_interface_num, var.environment)}"
+    desired_capacity = "${lookup(var.es_interface_num, var.environment)}"
     load_balancers = [
-        "elb-${var.environment}-socorroelasticsearch"
+        "elb-${var.environment}-socorroelasticsearch2"
```
rhelmer: One thing to note about names is that they can only be a max of 32 characters - `socorroelasticsearch` is quite a mouthful, maybe we should abbreviate to `socorro-es` or so?

rhelmer: I only bring this up because `elb-prod-socorroelasticsearch2` is 31 characters, according to `wc` :)

puppet/modules/socorro/templates/etc_elasticsearch/elasticsearch.yml.erb
phrawzty: Correct.  We are free to rename it, and if we want to, now is the perfect time to do so - ideas?

terraform/elasticsearch/main.tf
phrawzty: We could shorten the name but using a dash would violate the syntax that we appear to have established (and IIRC we can't use an underscore because that breaks some AWS conventions). Perhaps just "elasticsearch" (we know it's for Socorro in any case).

jdotpz: > we know it's for Socorro in any case
That is the case today, but in a world where we're hosting more than socorro on AWS, we want to be more specific so in the future we see "sumoelasticsearch" or "whateverappelasticsearch".  
Maybe socorroes ?

phrawzty: @jdotpz raises a solid point:
```
08:07:15 < jp> wrt naming: "we know it's for Socorro in any case". Well, today yeah, but if we
               add a bunch more apps?
08:07:39 < phrawzty> jp: true
```

jdotpz:
Enthusiastic r+

jdotpz:
dance

PR #134, opened 2015-05-22
fix comment and tag

rhelmer:
r? @phrawzty


phrawzty:
As long as we're cleaning, could you also fix [this typo](https://github.com/mozilla/socorro-infra/blob/master/terraform/collector/main.tf#L9)? :icecream:

jdotpz:
screenshot 2015-05-22 15 58 01

PR #135, opened 2015-05-22
Roll new variable scaling standard out to all active node types (+ a modification)

jdotpz:
## Highlights
* This PR applies @phrawzty 's new standard for scale variables to all other applicable node types (I didn't do the postgres nodes.)  
* It changes that standard a tiny bit with regards to max size management.
* Wait 15 times for the health checks to come back
* Deploy script now dabbles with min size to help stop losing nodes during longer deploys, and is cognizant of an apps current min_size when depoying.

## Changes to the previous standard
This also modifies https://github.com/mozilla/socorro-infra/pull/132#event-311899729 slightly.  The min/desired/max were all being set to the same variable.  Instead, I propose the following:

We make a semi-default max capacity; here, I chose 10 max for nodes we know won't really ever scale, and 30 for nodes that well...just might. I'm thinking if we need to change max, we'll need to do it on a more one-off basis, and can easily change the var in our repo per config file.

## Assumptions made
I took my instance type and counts from https://docs.google.com/spreadsheets/d/1TG0GSQK3RxoZCpy4DVOos5BomXZgD0LV0_L0dDrs7wQ/edit#gid=0 where we've specced out the first draft of what our infra will look like.

### DIFF
--- a/bin/deploy-socorro.sh
+++ b/bin/deploy-socorro.sh
@@ -104,7 +104,6 @@ function parse_github_payload() {
     # We will get all sorts of fun info here!
     echo "Git Payload for copying into jenkins build parameter"
     echo " ";echo ${GITPAYLOAD};echo " "
-    echo "Git Payload Readable: ${GITPAYLOAD}"|sed 's/,/\'$'\n/g';format_logs
     GITCOMMITHASH=$(echo ${GITPAYLOAD} | sed 's/,/\'$'\n/g'| grep head_commit | \
                     sed 's/"/ /g' | awk '{print $5}')
     echo "Git commit hash:  ${GITCOMMITHASH}"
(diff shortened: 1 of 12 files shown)


phrawzty:
With the introduction of env-based variables for scaling, instance types, etc., our `variables.tf` is going to get really bloated. The last time this happened (`main.tf`) we split out into a bunch of different sub-dirs for each role.  Might we do the same for variables?  Terraform will load every `*.tf` file in a given directory, so we could have a global file (like right now, symlinked), and then role-specific ones as well.

Taking #132 as an example, the variables contained in the Elasticsearch block would move out of `terraform/variables.tf` and into `terraform/elasticsearch/locals.tf`.  This helps to keep `main.tf` tidy and aligns nicely with the role-based model that we're already using.

Thoughts? @jdotpz @rhelmer

phrawzty:
@jdotpz and I discussed the splitting out of `variables.tf` - for now we'll keep it as a monolithic file since this PR has a lot more functionality that we'd like to roll out. We can split the variables up in a later PR.

jdotpz:
Woooooo.

phrawzty:
thumbsup

PR #137, opened 2015-05-27
bug 1118471; es data nodes

phrawzty:
This ensures that the ES Data nodes have an EBS volume available to them. This is done in two parts: in Terraform an `ebs_block_device` resource is added to the appropriate launch config, and in Puppet the volume is formatted and mounted before the ES service is activated.

I also made some style fixes in both Puppet and Terraform which are cosmetic (and make the PR look larger than it is).

`r?` @jdotpz @rhelmer

### DIFF
--- a/puppet/modules/socorro/manifests/role/elasticsearch.pp
+++ b/puppet/modules/socorro/manifests/role/elasticsearch.pp
@@ -4,11 +4,12 @@ class socorro::role::elasticsearch {
 include socorro::role::common
 
   # ES will hit default ulimits rather quickly.
-  file { '/etc/security/limits.d/90-elasticsearch.conf':
-    owner  => 'root',
-    group  => 'root',
-    mode   => '0644',
-    source => 'puppet:///modules/socorro/etc_security_limits.d/90-elasticsearch.conf'
+  file {
+    '/etc/security/limits.d/90-elasticsearch.conf':
+      owner  => 'root',
+      group  => 'root',
+      mode   => '0644',
+      source => 'puppet:///modules/socorro/etc_security_limits.d/90-elasticsearch.conf'
   }
 
   # These switches determine the role of the node: master, interface, or data.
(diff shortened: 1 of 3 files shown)


terraform/variables.tf
```diff
@@ -67,8 +67,8 @@ variable "es_interface_num" {
 }
 variable "es_data_ec2_type" {
     default = {
-        stage = "i2.xlarge"
-        prod = "i2.2xlarge"
+        stage = "r3.xlarge"
```
phrawzty: I changed these to the `r` type since we are no longer concerned with instance storage (these cost less, too). :money_with_wings:

puppet/modules/socorro/manifests/role/elasticsearch.pp
```diff
@@ -26,11 +27,17 @@
...
-    owner   => 'root',
-    group   => 'elasticsearch',
-    mode    => '0644',
-    content => template('socorro/etc_elasticsearch/elasticsearch.yml.erb')
+  file {
+    '/etc/elasticsearch/elasticsearch.yml':
+      owner   => 'root',
+      group   => 'elasticsearch',
+      mode    => '0644',
+      content => template('socorro/etc_elasticsearch/elasticsearch.yml.erb')
```
jdotpz: Maybe a notify => Service['elasticsearch'] here to be sure it drops the config in and the process runs with that config?

puppet/modules/socorro/manifests/role/elasticsearch.pp
```diff
@@ -40,8 +47,32 @@
...
+      '/var/lib/elasticsearch':
+        ensure  => mounted,
+        device  => '/dev/xvdb',
+        fstype  => 'ext4',
+        options => 'defaults',
+        require => Exec['format-es-data']
+    }
+
+    File['/var/lib/elasticsearch'] {
+      require => Mount['/var/lib/elasticsearch']
```
jdotpz: Maybe a notify here?

puppet/modules/socorro/manifests/role/elasticsearch.pp
phrawzty: Note that the ES service is *not* running by this point (that's what the service resource does). Said service has requires on those resources and won't trigger without them, so there's nothing to notify.

jdotpz:
Once it tests out well, a hearty r+

phrawzty:
> Once it tests out well, a hearty r+

Terraform apply is good! Nodes come up, partitions are mounted, and the cluster is green.

PR #139, opened 2015-05-28
fix bug 1168723; allow Consul for all nodes

phrawzty:
This PR adds the same same two ingress rules to all of the EC2-based roles (except `consul` which already had them, heh). These rules allow all of the nodes to communicate with the Consul raft freely.

`r?` @jdotpz @rhelmer


rhelmer:
r+

PR #140, opened 2015-05-28
set hostname before joining consul

phrawzty:
Ensure that the hostname has been set prior to joining the Consul cluster (this addresses a minor irritation more than anything, heh).

We do need to address the issue of when (and how) to set the hostname in a more general way. @jdotpz suggested doing it during cloud-init which makes sense to me. In any case, this PR can be merged while we work on a better solution.

`r?` @jdotpz @rhelmer


rhelmer:
r+

PR #144, opened 2015-05-29
Add base datadog agent to all nodes

jdotpz:
I'm closing old https://github.com/mozilla/socorro-infra/pull/128 in favor of a cleaner pull request for datadog agent.

* In packer phase, via socorro base, we setup and enable the yum repo for datadog agent as well as install the rpm (unsigned, btw)

* In role phase, via common.pp, we give each node a templated datadog agent config file (in /etc/dd-agent/datadog.conf) and bounce the datadog agent service.  For the moment, the only thing I've changed in this datadog.conf is our API key (get it from hiera) and I've enabled EC2 tags collection so we can associate our role/environment/project tags with these agents.  Moving forward, it's likely we template/variablize a bunch of other stuff in the datadog.conf file (like logging to syslog, log parsers, etc).

### DIFF
--- a/puppet/modules/socorro/manifests/packer/base.pp
+++ b/puppet/modules/socorro/manifests/packer/base.pp
@@ -14,6 +14,11 @@ class socorro::packer::base {
       enable  => false,
       require => Package['postgresql93-server'];
 
+    'datadog-agent':
+      ensure  => stopped,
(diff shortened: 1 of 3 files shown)


rhelmer:
r+

PR #145, opened 2015-05-30
according to the consul docs, we need UDP egress from all nodes

rhelmer:
See https://www.consul.io/docs/agent/options.html "Ports Used"

r? @jdotpz


jdotpz:
A +1 contingent in later locking it down a bit more.  Let's get the infra up.

jdotpz:
https://bugzilla.mozilla.org/show_bug.cgi?id=1167722 has us tidying a bunch of this type of thing up pre-migration.

PR #146, opened 2015-05-31
re-add small symbol cache, no simple way to disable this in stackwalk…

rhelmer:
…er yet

r? @jdotpz @phrawzty


jdotpz:
Good to go.

PR #152, opened 2015-06-02
workaround bug 1170420 - ensure symbols cache dir exists, until this …

rhelmer:
…is fixed in Socorro Processor

r? @jdotpz


rhelmer:
#153 fixes this.

PR #154, opened 2015-06-02
Add more instrumentation to datadog in elasticsearch / rabbitmq

jdotpz:
## Knowing things is awesome!

Here we add elasticsearch and rabbitmq datadog agents.
Since we do an include, I believe we can double define resources (such as service['datadog-agent'])

### DIFF
--- a/puppet/modules/socorro/files/etc_dd-agent/elasticsearch.yaml
+++ b/puppet/modules/socorro/files/etc_dd-agent/elasticsearch.yaml
@@ -1,0 +1,22 @@
+init_config:
+
+instances:
+  # The URL where elasticsearch accepts HTTP requests. This will be used to
+  # fetch statistics from the nodes and information about the cluster health.
+  #
+  # If you're using basic authentication with a 3rd party library, for example
+  # elasticsearch-http-basic, you will need to specify a value for username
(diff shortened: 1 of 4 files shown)


rhelmer:
r+

puppet/modules/socorro/files/etc_dd-agent/elasticsearch.yaml
```diff
@@ -0,0 +1,22 @@
...
+  # If you're using basic authentication with a 3rd party library, for example
+  # elasticsearch-http-basic, you will need to specify a value for username
+  # and password for every instance that requires authentication.
+  #
+  # If your cluster is hosted externally (i.e., you're not pointing to localhost)
+  # you will need to set `is_external` to true otherwise the check will compare
+  # the local hostname against hostnames of the Elasticsearch nodes and only
+  # submit metrics if they match.
+  #
+  - url: http://localhost:9200
```
phrawzty: Not sure if it matters, but this will only work on the *interface* types, since the other two types don't have HTTP endpoints.

phrawzty: Ah, nevermind - it's only applied in the context of the interface role. Good good.

PR #156, opened 2015-06-02
bug 1170612 - use more standard /mnt for symbol cache mountpoint

rhelmer:
r? @jdotpz - I tested this on stage infra and Really Werks!

### DIFF
--- a/puppet/modules/socorro/manifests/role/processor.pp
+++ b/puppet/modules/socorro/manifests/role/processor.pp
@@ -10,7 +10,7 @@ include socorro::role::common
   }
 
   mount {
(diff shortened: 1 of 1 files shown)


puppet/modules/socorro/manifests/role/processor.pp
```diff
@@ -10,7 +10,7 @@
   }
 
   mount {
-    '/tmp/symbols':
+    '/mnt':
```
phrawzty: I'm not sure how I feel about mounting at `/mnt` directly. Should we ever want to mount more than one thing this may bite us in the ass later. :bug:

PR #163, opened 2015-06-04
Cleanup which roleenvs get updates

jdotpz:
This cleans up elasticsearch infra from the nodes we update with special settings (ie, elb alarms, etc, all from updateinfrastructure script).
It also fixes a bug in the Jenkins backup job.

### DIFF
--- a/bin/jenkins/backup_jenkins.sh
+++ b/bin/jenkins/backup_jenkins.sh
@@ -58,10 +58,10 @@ function wait_for_ami() {
     PROGSTEP="Waiting for AMI"
     echo "`date` -- Waiting 300 seconds for ${AMIID} to become available"
     sleep 300
-    until aws ec2 describe-images --image-id ${AMIID} --output text --query 'Images[*}.State'| grep available > /dev/null;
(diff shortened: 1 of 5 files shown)


jdotpz:
This also now fixes ES and rmq's double declaration.

bin/jenkins/backup_jenkins.sh
```diff
@@ -58,10 +58,10 @@ function wait_for_ami() {
     PROGSTEP="Waiting for AMI"
     echo "`date` -- Waiting 300 seconds for ${AMIID} to become available"
     sleep 300
-    until aws ec2 describe-images --image-id ${AMIID} --output text --query 'Images[*}.State'| grep available > /dev/null;
+    until aws ec2 describe-images --image-id ${AMIID} --output text | grep available > /dev/null
         do
         echo "`date` -- Waiting for ${AMIID} to become available"
-        sleep 30
+        sleep 60
```
phrawzty: Hardcoded `sleep`s make me twitchy (the `sleep 300` above, too). Is there a way we could query for a condition instead of just waiting blindly?

PR #165, opened 2015-06-08
we use PG 9.4 on RDS. Also remove plperl, we're not using that

rhelmer:
r? @jdotpz


lonnen:
r+ when merge queue has died down a little

PR #166, opened 2015-06-09
fix bug 1172892; swap in emergencies only

phrawzty:
The [ES docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/setup-configuration.html) state that swap should be disabled (or at least reduced) and suggest three ways to accomplish this goal. Of the three, I prefer using sysctl to adjust vm.swappiness since this is straightforward to do via Puppet.

@rhelmer @jdotpz `r?`


rhelmer:
r+

PR #167, opened 2015-06-09
bug 1170204 - further adjust ES heap sizes to fit current instance types

rhelmer:
r? @phrawzty

### DIFF
--- a/puppet/modules/socorro/files/etc_sysconfig/elasticsearch.erb
+++ b/puppet/modules/socorro/files/etc_sysconfig/elasticsearch.erb
@@ -3,11 +3,11 @@ ES_HOME=/usr/share/elasticsearch
 
 # Heap Size (defaults to 256m min, 1g max)
 <% if @es_master %>
-ES_HEAP_SIZE=24g
+ES_HEAP_SIZE=15g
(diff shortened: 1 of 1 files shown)


phrawzty:
Within the context of ES, heap sizes are governed by [two factors](https://www.elastic.co/guide/en/elasticsearch/guide/current/heap-sizing.html): Lucene and Java.

Lucene operates out of the heap (it's functionally filesystem operations), and therefore has separate memory requirements. As a general rule, one should never allocate more than 50% of the total system RAM to the ES Java heap - the other half is left to the kernel to be used for the system, which includes Lucene.

Java catches a bad case of *the dumb* when it has more than 32GB of memory to deal with. Allocating more than that to the heap becomes a question of diminishing returns, and can even have detrimental effects for smaller allocation sizes; unless one is allocating *hundreds* of gigs it's not even worth dealing with the issue.

In summary, these numbers should not exceed 32GB, and must not exceed 50% of total system.

phrawzty:
`r+`

As an aside, there is a potential argument to be made for violating the "50%-rule" on the *master* nodes since they aren't actually doing any heavy lifting with Lucene. Something to think about for the future, perhaps.

PR #168, opened 2015-06-09
Auto-backups for consul

jdotpz:
Here, we move install of pip out of buildbox and into basebox, since our consul nodes will want pip, awscli, and consulate installed on them with the pip provider.

### DIFF
--- a/puppet/modules/socorro/manifests/packer/base.pp
+++ b/puppet/modules/socorro/manifests/packer/base.pp
@@ -66,6 +66,7 @@ class socorro::packer::base {
       'mod_wsgi',
       'nginx',
       'php-cli',
+      'python-pip',
       'rabbitmq-server',
       'supervisor',
       'unzip',
@@ -104,6 +105,7 @@ class socorro::packer::base {
       require => Package['elasticsearch']
   }
 
+
   file {
     '/etc/socorro':
       ensure => directory;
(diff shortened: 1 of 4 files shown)


rhelmer:
r+

puppet/modules/socorro/manifests/packer/base.pp
```diff
@@ -104,6 +105,7 @@
       require => Package['elasticsearch']
   }
 
+
```
phrawzty: Whitespace?

puppet/modules/socorro/manifests/packer/base.pp
```diff
@@ -123,6 +125,31 @@
       notify  => Service['elasticsearch'];
   }
 
+
```
phrawzty: Whitespace?

phrawzty:
How resource-intensive is the key/value dump? Does it block consul at all? If so, there may be value in adding a splay to this such that the hosts run their backups at different times from one another...

rhelmer:
@phrawzty following https://github.com/gmr/consulate/blob/dd84c0472f35d9eb6ebe71fd7e8ade4cd20cfc5a/consulate/cli.py#L151 through the code, I'd say it's probably not very resource intensive... it seems to just be using the Consul HTTP API to get a list of key/value pairs and dumping that.

(the above ultimately boils down to https://github.com/gmr/consulate/blob/938f96ca3dab4b2d9ad42a1aa76e702e1be00802/consulate/api/base.py#L61-L63 with `adapter` being the Python `requests` module by default)

rhelmer:
I think this is good to go, splay should not be necessary.

PR #169, opened 2015-06-11
bug 1126369; add more analysis packages

phrawzty:
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1126369#c5 .

`r?` @rhelmer

### DIFF
--- a/puppet/modules/socorro/manifests/role/analysis.pp
+++ b/puppet/modules/socorro/manifests/role/analysis.pp
@@ -30,6 +30,10 @@ include socorro::role::common
   package {
     [
       'php-cli',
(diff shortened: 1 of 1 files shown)


rhelmer:
r+ thanks for grabbing those!

puppet/modules/socorro/manifests/role/analysis.pp
```diff
@@ -30,6 +30,10 @@
   package {
     [
       'php-cli',
+      'php-pgsql',
+      'php-xml',
+      'mercurial',
+      'nano',       # really?
```
rhelmer: lulz

PR #170, opened 2015-06-11
[bug 1173085] deal with ebs mount edge case

phrawzty:
Since cloud-init is a bit of an unknown for us right now, I reasoned that Puppet presented the most rapid path to dealing with the EBS mount [edge case](https://bugzilla.mozilla.org/show_bug.cgi?id=1173085).

I've tested this and, from a system perspective, it appears to work as expected. :grin:  In terms of actual deployment there are some configuration (Consul) elements that will need to be adjusted.  Let's open a dialogue on the best way to roll this out.

`r?` @rhelmer @jdotpz

### DIFF
--- a/puppet/modules/socorro/manifests/role/processor.pp
+++ b/puppet/modules/socorro/manifests/role/processor.pp
@@ -1,28 +1,55 @@
 # Set up a processor node.
 class socorro::role::processor {
 
-include socorro::role::common
+  include socorro::role::common
+
+  # Symbols uses a hierarchy of directories that should be on the same
+  # block device for performance reasons.
+  $symbols_base = '/mnt/symbols'
+  $symbols_dirs = [
+    "${symbols_base}/cache",
+    "${symbols_base}/tmp"
+  ]
+
+  file {
+    $symbols_base:
+      ensure => directory
+  }
+
+  # There is an edge-case interaction between cloud-init and *some* instance
+  # types that causes EBS volumes to be pre-mounted.
+  # https://bugzilla.mozilla.org/show_bug.cgi?id=1173085
+  exec {
+    'check-premounted-ebs':
+      path    => '/bin',
+      command => 'umount /dev/xvdb',
+      onlyif  => 'mount | grep xvdb'
+  }
 
   exec {
-    'format-symbol-cache':
+    'format-symbols-cache':
       path    => '/usr/sbin',
-      command => 'mkfs.ext4 /dev/xvdc'
+      command => 'mkfs.ext4 /dev/xvdb',
+      require => Exec['check-premounted-ebs']
   }
 
   mount {
-    '/mnt':
+    $symbols_base:
       ensure  => mounted,
-      device  => '/dev/xvdc',
+      device  => '/dev/xvdb',
       fstype  => 'ext4',
       options => 'defaults',
-      require => Exec['format-symbol-cache']
+      require => [
+        Exec['format-symbols-cache'],
+        File[$symbols_base]
+      ]
   }
 
   file {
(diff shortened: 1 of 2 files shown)


terraform/processor/main.tf
```diff
@@ -94,18 +93,18 @@ resource "aws_autoscaling_group" "as-processor" {
     min_size = "${lookup(var.processor_num, var.environment)}"
     desired_capacity = "${lookup(var.processor_num, var.environment)}"
     tag {
-      key = "Environment"
-      value = "${var.environment}"
-      propagate_at_launch = true
+        key = "Environment"
```
phrawzty: Bonus: I fixed some whitespace issues. :smirk:

terraform/processor/main.tf
```diff
@@ -71,9 +71,8 @@ resource "aws_launch_configuration" "lc-processor" {
     lifecycle {
         create_before_destroy = true
     }
-
     ephemeral_block_device {
-        device_name = "/dev/xvdc"
+        device_name = "/dev/xvdb"
```
phrawzty: I set this to **`b`** since it's the second device; if it needs to be `c` we can switch back.

puppet/modules/socorro/manifests/role/processor.pp
```diff
@@ -1,28 +1,55 @@
...
+
+  file {
+    $symbols_base:
+      ensure => directory
+  }
+
+  # There is an edge-case interaction between cloud-init and *some* instance
+  # types that causes EBS volumes to be pre-mounted.
+  # https://bugzilla.mozilla.org/show_bug.cgi?id=1173085
   exec {
```
phrawzty: This is a simple and decidedly inelegant solution: if the mount is there, unmount it. We can get away with this because we're using a strict role separation model; however, if we ever intend to combine roles, this *may* be problematic down the line.

puppet/modules/socorro/manifests/role/processor.pp
```diff
@@ -1,28 +1,55 @@
 # Set up a processor node.
 class socorro::role::processor {
 
-include socorro::role::common
+  include socorro::role::common
 
+  # Symbols uses a hierarchy of directories that should be on the same
+  # block device for performance reasons.
```
rhelmer: I think this comment is true enough, I just wanted to point out that if you don't put these on the same device stackwalker will totally fail to cache symbols at all (since the `rename()` will fail).

It's true that the reason it does `rename()` instead of a "copy-and-delete" is for perf reasons though, so true enough :)

puppet/modules/socorro/manifests/role/processor.pp
rhelmer: I'm not sure what you mean by combine roles, like apply multiple roles to the same node?

rhelmer:
r+, in terms of rollout it's totally fine to just let this go out and configure processor after the fact - incoming jobs will simply be queued up. This is preventing new processor instances from coming up cleanly anyway.

puppet/modules/socorro/manifests/role/processor.pp
phrawzty: I mean if we combine two roles that both expect exclusive use of `/dev/xvdb`. Unlikely, but wanted to point it out anyway.

rhelmer: Ah good point - yes I would definitely like to handle ephemeral storage in a more general way. I think this is fine for the moment since it's only really critical for processor.

phrawzty: The ES data nodes also mount this device, fwiw...

PR #173, opened 2015-06-15
Lock down ingress rules for all role types

jdotpz:


terraform/admin/main.tf
```diff
@@ -7,12 +7,32 @@ provider "aws" {
...
     name = "ec2-socorroadmin-${var.environment}-sg"
     description = "Allow (alt) SSH to the Admin node."
+    # phrawzty home
     ingress {
         from_port = "${var.alt_ssh_port}"
         to_port = "${var.alt_ssh_port}"
         protocol = "tcp"
         cidr_blocks = [
-            "0.0.0.0/0"
+            "${var.phrawzty_ip}"
```
rhelmer: @jdotpz do you think we should be this specific? Maybe we could just have a list of `admin` IPs? I'm not sure if we can apply a list to a single ingress block though...

phrawzty: I'm with @rhelmer on this one - this is a bad smell, but to be honest, I'm not sure what a better solution looks like.  What would be really nice is an array primitive that we could just import. :camel:

terraform/analysis/main.tf
```diff
@@ -7,12 +7,32 @@ provider "aws" {
...
     name = "ec2-socorroanalysis-${var.environment}-sg"
     description = "Crashanalysis node."
+    # phrawzty home
     ingress {
         from_port = "${var.alt_ssh_port}"
         to_port = "${var.alt_ssh_port}"
         protocol = "tcp"
         cidr_blocks = [
-            "0.0.0.0/0"
+            "81.64.0.0/16"
```
phrawzty: In principle this shouldn't be my block - it should be a reference to the variable.

Same for all remaining `.tf`'s below.

jdotpz: Ah, I must have messed up there, apologies!  I'll get that variable in there.

terraform/admin/main.tf
jdotpz: For if we ever need to handle PII or PCI or any other compliance-requiring stuffs, it's later much easier to so if you have things pretty specified to individuals so you can show that you are maintaining that list diligently.  That was my thinking on putting in specific by-person rules.

terraform/analysis/main.tf
jdotpz: Fixed and squashed.

phrawzty:
I'm working from a diverse range of remote locations for the next two weeks (true story) - what's my solution for access in this model?

jdotpz:
You can go manual and add your current IP to the security group, as rhelmer was mentioning earlier.  Know that if you're working on one of the nodes we update (collector, webapp, and processor atm), and a push goes out, your manual rule change will need to be remade.

terraform/admin/main.tf
```diff
@@ -48,6 +68,7 @@ resource "aws_security_group" "ec2-socorroadmin-sg" {
             "172.31.0.0/16"
         ]
     }
+
```
phrawzty: Whitespace.

phrawzty:
`r+` from me, but I'd like to get @rhelmer to sign off on this as well.

rhelmer:
r+

phrawzty:
MERGE CONFLICT 2015

jdotpz:
Who else do we need to add IP addresses for?

I know lars has been accessing things, any others?  I'll add lars and $whoeverelse and clean up those nasty, hateful conflicts.

phrawzty:
> I know lars has been accessing things, any others? I'll add lars and $whoeverelse and clean up those nasty, hateful conflicts.

What about the Mozilla offices themselves?

jdotpz:
Good call, I'll get those added.  I got Lars' ip block, and will update this pr with his and that.  Any others?  cc @rhelmer

rhelmer:
@jdotpz 

> Good call, I'll get those added. I got Lars' ip block, and will update this pr with his and that. Any others? cc @rhelmer

* vladan and wcosta have access to symbolapi
* kairo has access to socorroanalysis

PR #175, opened 2015-06-15
info on bootstrapping services

rhelmer:
r? @jdotpz @phrawzty - not saying we shouldn't further automate bringing up our long-running stateful services, but I feel like for the moment they at least need to be more clearly documented.

This replaces PR #49


jdotpz:
r+ beautiful, will merge as soon as travis +'s it.

PR #178, opened 2015-06-16
Variables for separate prod and stage certs

jdotpz:


terraform/variables.tf
```diff
@@ -3,10 +3,23 @@ variable "access_key" {}
 variable "secret_key" {}
 variable "secret_bucket" {}
 variable "subnets" {}
-variable "collector_cert" {}
-variable "webapp_cert" {}
-variable "analysis_cert" {}
+variable "collector_cert" {
```
phrawzty: Proper format for map variables is:
```
variable "var_name" {
    default = {
        this = "THIS"
        that = "THAT"
    }
}
```

That default block is important - the map is invalid without it. Irritatingly, Terraform doesn't appear to warn/error on it directly though, which is why we didn't catch it right away, heh.

terraform/variables.tf
```diff
@@ -3,10 +3,31 @@ variable "access_key" {}
...
+    default = {
+        prod  = "fakevalue"
```
phrawzty: For purely aesthetic reasons, mightn't it be nicer to just have this as empty vars instead of `fakevalue`? :ideograph_advantage:

phrawzty:
`test.sh` will need to be updated such that the certs which were previously scalars are dealt with appropriately. Since we're declaring them as maps with empty (but defined) values, we can likely just remove them from `test.sh` entirely.

Also, upgrade the TF ver to 0.5.3 while we're at it.

phrawzty:
`r+` on ab51a15.

PR #179, opened 2015-06-18
launch config roles should not attempt package upgrades

rhelmer:
r? @jdotpz - we want these packages to be baked into the AMI, and not modified at launch time. In particular, the Socorro package doesn't always upgrade cleanly so could cause problems (that's how I noticed this, on the webapp :) )


jdotpz:
r+, gj snaking this out.

PR #180, opened 2015-06-22
bug 1176673 - bump idle timeout for symbol upload

rhelmer:
r? @jdotpz / @phrawzty


phrawzty:
The diff is fine, but I'm curious about the value: why 600 seconds exactly?  Arbitrary or scientific? :smile:

rhelmer:
@phrawzty per the bug, it looks like uploads are timing out around 3 minutes, so I meant to make this 300s. Not scientific per se just a nice round number.

rhelmer:
@phrawzty rebased and pushed 300s

jdotpz:
r+

PR #181, opened 2015-06-23
Adjust elb timeout

rhelmer:
r? @phrawzty / @jdotpz - sorry, re-reading the bug, I missed a few things:

* both collector and webapp ELBs should have 5 minute timeout (from default 60s)
* nginx idle timeout should be 5 minutes (default is 60s as well)


jdotpz:
r+, travis likes it, shipping it.

jdotpz:
It's on builder now.

PR #182, opened 2015-07-06
Fix errors in update-infrastructure.sh to correctly setup autoscaling…

jdotpz:
… policies and alarms


jdotpz:
This fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1108279

phrawzty:
`r+` on 4877d4a; however, what's the deal with the cipher script? Do we have a bug open on that?

PR #183, opened 2015-07-06
Resolve https://bugzilla.mozilla.org/show_bug.cgi?id=1170273

jdotpz:
The idea here is to retry when we see 'Traceback' in the temp log created during the package build.

### DIFF
--- a/bin/lib/create_rpm.sh
+++ b/bin/lib/create_rpm.sh
@@ -18,9 +18,26 @@ function create_rpm() {
         # Get the hash for later use to feed packer
         export SOCORROHASH=`git log | head -n1 | awk '{print $2}'`
         echo "`date` -- Clone of commit for ${SOCORROHASH} returned ${RETURNCODE}"
+        CONTINUE="no"
+        until [ "${CONTINUE}" = "yes" ];do
         # Build the actual rpm file
-        /usr/bin/env PYTHON=python make package BUILD_TYPE=rpm
-            RETURNCODE=$?;error_check
+        set -o pipefail
+        /usr/bin/env PYTHON=python make package BUILD_TYPE=rpm | tee /var/log/${RANDOM_STRING}
+            RETURNCODE=$?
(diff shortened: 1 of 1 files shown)


bin/lib/create_rpm.sh
```diff
@@ -18,9 +18,26 @@ function create_rpm() {
         # Get the hash for later use to feed packer
         export SOCORROHASH=`git log | head -n1 | awk '{print $2}'`
         echo "`date` -- Clone of commit for ${SOCORROHASH} returned ${RETURNCODE}"
+        CONTINUE="no"
+        until [ "${CONTINUE}" = "yes" ];do
```
phrawzty: Is there a specific reason that you're testing the expansion instead of the value? It's a subtle difference but I feel the latter is more standard / readable.
```bash
until [ $CONTINUE == "yes" ]; do
```
Also, be careful with the comparison vs. assignment operators.

bin/lib/create_rpm.sh
```diff
@@ -18,9 +18,26 @@ function create_rpm() {
...
         # Build the actual rpm file
-        /usr/bin/env PYTHON=python make package BUILD_TYPE=rpm
-            RETURNCODE=$?;error_check
+        set -o pipefail
```
phrawzty: This is all inside of the `do` block that starts on line 22 and thus needs to be indented over.

Where does this block end? I.e. where is the `done`?

phrawzty:
It is conceivable that this could result in an infinite loop. Imagine a scenario where a non-fatal traceback happens over and over - this script would never stop grinding. There should absolutely be some sort of fail-safe to prevent this behaviour from occurring (a counter that just terms after a set number of tries, for example).

bin/lib/create_rpm.sh
jdotpz: Eeek!  Good catch!

jdotpz:
Yeah, agreed on the infinite loop bit.  As I look back on this PR, I dislike it more and more.  Let's kill this.

PR #184, opened 2015-07-07
[bug 1181132] Adjust reporting for new AMI

jdotpz:
This moves the beginning report to before the new ami is applied.  Duh.


phrawzty:
`r+`

PR #186, opened 2015-07-16
Further lock down ES SG for 9300, addressing https://bugzilla.mozilla…

jdotpz:
….org/show_bug.cgi?id=1184706


jdotpz:
Hmm, so this doesn't sound like it's going to be addressed by Terraform, unfortunately.  We should probably update the setup docs with instructions to manually add this to any new ES clusters we build.

lonnen:
Newer versions of terraform have a `self` param for `ingress` and `egress` blocks. We would need to upgrade to get this to work. Recommend wontfix.

PR #187, opened 2015-07-20
Bug1185680 check in variables and user data

rhelmer:
r? @jdotpz - this makes our terraform config work with Atlas.


phrawzty:
Having to put `socorro_role.sh` everywhere is so gross, but it doesn't appear that we can avoid it if we want to want to stick with our isolation model.

Regarding the Travis failure on 273375f, you'll need to adjust line 14 of `test.sh` to stop the symlinks from getting created. I suppose that entire bit of functionality could be removed from `wrapper.sh` ultimately.

jdotpz:
Gah sorry @rhelmer, I let this sit and now it's got conflicts.

jdotpz:
@rhelmer if you can clean up the conflicts, I'll merge this and get it on builder.

rhelmer:
@jdotpz rebased against master

PR #188, opened 2015-07-21
Fixes bug 1175946 - Enabled Groovy scripting in our Elasticsearch clu…

AdrianGaudebert:
…sters.

@jp or @phrawzty r?

### DIFF
--- a/puppet/modules/socorro/files/etc_elasticsearch/elasticsearch.yml.erb
+++ b/puppet/modules/socorro/files/etc_elasticsearch/elasticsearch.yml.erb
@@ -23,3 +23,5 @@ node:
     data: <%= @es_data %>
 http:
     enabled: <%= @es_interface %>
+script:
+    disable_dynamic: false


AdrianGaudebert:
I have no idea how to test this though... I mean I know what to do to verify it solves my problem, and I have verified locally that this setting is the one, but I don't know how to specifically test this ``erb`` file. I suppose stage is the way to go?

phrawzty:
> I suppose stage is the way to go? 

wisely

jdotpz:
@phrawzty would you recommend just doing this manually on all existing nodes (stage first, as wisely chosen), or would you recommend bringing in new fresh nodes built from a new AMI with this manifest?

phrawzty:
> @phrawzty would you recommend just doing this manually on all existing nodes (stage first, as wisely chosen), or would you recommend bringing in new fresh nodes built from a new AMI with this manifest?

Given how irritating it is to replace ES nodes, I would just apply the setting manually via an interface node.

AdrianGaudebert:
@jdotpz so can we do this?

jdotpz:
Sure.  @phrawzty apologies to be an ES noob...is that documented somewhere on the process to use to apply those settings to the whole cluster?

phrawzty:
> Sure. @phrawzty apologies to be an ES noob...is that documented somewhere on the process to use to apply those settings to the whole cluster?

You just need to submit the config change via an HTTP interface as normal - it will be propagated to the rest of the cluster without further intervention.

jdotpz:
Have a link to docs showing how?  I can look at configs and such to figure out parameter names.

phrawzty:
> Have a link to docs showing how? I can look at configs and such to figure out parameter names.

The scripting engine (and config) has evolved quite a bit between [1.4](https://www.elastic.co/guide/en/elasticsearch/reference/1.4/modules-scripting.html) and 1.7(https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting.html) (current), so you'll really need to review the docs that pertain to the version that's currently deployed.

That said, upon further reflection, it's possible that this isn't a setting that can be applied dynamically (i.e. via POSTing to the interface), in which case the setting will need to be applied to the configs, followed by a rolling reload.

jdotpz:
Sure, that I can do, and check cluster health during.  I'm going to wait until my irc client works before doing it, though.

jdotpz:
I've manually updated stage elasticsearch cluster with this config, and in a gentle rolling fashion, recycled the process on the servers.

We're back to green health:
{
  "cluster_name" : "sss_stage",
  "status" : "green",
  "timed_out" : false,
  "number_of_nodes" : 8,
  "number_of_data_nodes" : 3,
  "active_primary_shards" : 45,
  "active_shards" : 90,
  "relocating_shards" : 2,
  "initializing_shards" : 0,
  "unassigned_shards" : 0
}

So we test this for a bit in stage, and then we do the same on prod.

AdrianGaudebert:
I cannot fully verify that this works correctly without having https://github.com/mozilla/socorro-infra/pull/195 merged first. Though it seems the error I just got was a timeout and not an Elasticsearch error, I'd rather be sure. :)

jdotpz:
Merged 195, waiting for a staging push for it.

AdrianGaudebert:
So this actually cannot work, the setting I'm using is for ES 1.6+ and is not understood by ES 1.4. Discovered by talking with @phrawzty, confirmed with @peterbe in https://github.com/mozilla/socorro/pull/2926 and just now in sentry: https://errormill.mozilla.org/webtools/socorro-stage/group/399070/

Updating this PR in a few moments...

AdrianGaudebert:
Updated, @jdotpz can we try this again? Sorry about the mess!

peterbe:
I thought we had Groovy in production already. Didn't you need that to make certain fixes to super search fields which was resolved?

AdrianGaudebert:
@jdotpz since this is resolved in production now, should we merge this PR?

PR #191, opened 2015-07-28
Set hostname during socorro-role.sh to make logging better

jdotpz:
Instead of setting the hostname during puppet, we can set the hostname earlier in socorro-role.sh.  We thus will now it's there when puppet runs things like starting rsyslog/datadog agent.


puppet/modules/socorro/manifests/role/common.pp
```diff
@@ -25,18 +9,14 @@
       require => Exec['set-hostname']
   }
 
-  #  $logging_hostname = hiera("${::environment}/logging_hostname")
   file {
     '/etc/rsyslog.d/30-socorro.conf':
-      source  => 'puppet:///modules/socorro/etc_rsyslog/30-socorro.conf',
-      owner   => 'root',
+      source  => 'puppet:///modules/socorro/etc_rsyslog/30-socorro.conf',      owner   => 'root',
```
rhelmer: this is making `puppet-lint` grumpy

rhelmer:
r+ (when travis is happy)

PR #194, opened 2015-07-28
Restart rsyslog after applying hostname

jdotpz:
Noticed some of the hosts didn't start reporting as the new hostname until i manually restarted rsyslog.


rhelmer:
r+

PR #205, opened 2015-08-03
Bug 1188085 - Added new required Elasticsearch configuration to the w…

adngdb:
…ebapp.

@jdotpz r? cc @peterbe


AdrianGaudebert:
This is required before we can do a new release to production (otherwise all Super Search based pages would break, as they are breaking on stage).

jdotpz:
@AdrianGaudebert what do those values need to be set for?  This will actually require a live config change in Consul in addition to this PR.

peterbe:
@jdotpz We need to set environment variable `ELASTICSEARCH_URLS` to the same value that we're setting and giving the middleware app. 
I'm not familiar with what `@@@ES_URLS@@@` is at all or how that `@@@` stuff works. 
But check in consul for other uses of `ELASTICSEARCH_URLS` or something and we basically need to have that same value now present for the webapp.

rhelmer:
@peterbe the `@@@` things are really just placeholders, they don't get replaced automatically by anything: https://github.com/mozilla/socorro-infra/tree/master/socorro-config#replace-placeholders

It's just a hint that it needs configuration for your environment.

@jdotpz - IMHO should nuke this style of config and switch to `consulate` style.

jdotpz:
That is consulate,   The conf files are simply to fill in variables when we
load consul.  It's not intuitive, don't worry about worrying about it. :)

PR #206, opened 2015-08-03
Fix missing hostname updates in most roles

jdotpz:
When we added socorro_role.sh into each existing folder instead of symlinking, we lost our fancy new hostnames.


rhelmer:
r+

PR #207, opened 2015-08-03
Allow specific AMI ids to be used instead of hash for replace ami job

jdotpz:


jdotpz:
I'm about to use this...so, uh, merging.

PR #211, opened 2015-08-19
fixes bug 1196344 - Correctly cache static assets

peterbe:
@jdotpz r?

I'm not familiar with the `alias` command in nginx but I guess it's like a rewrite. Instead it's better to just set the root and any URL that can't be mapped to a file on the file system gets sent as a request to Django. 

This is the set up I have on my personal server.

### DIFF
--- a/puppet/modules/socorro/files/etc_nginx/conf_d/socorro-webapp.conf
+++ b/puppet/modules/socorro/files/etc_nginx/conf_d/socorro-webapp.conf
@@ -11,9 +11,18 @@ server {
     # crash-stats needs to accept debug symbol zips
     client_max_body_size 1g;
 
+    # all URLs under this prefix will have unique URLs
     location /static/CACHE {
         add_header   Cache-Control public;
         expires      max;
+        access_log   off;
+        break;
+    }
+
+    # files like /static/browserid.js that don't have a unique URL
+    location /static {
+        add_header   Cache-Control public;
+        expires      1d;
         access_log   off;
         break;
     }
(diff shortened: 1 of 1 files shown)


puppet/modules/socorro/files/etc_nginx/conf_d/socorro-webapp.conf
```diff
@@ -2,6 +2,8 @@ server {
     listen 80 default_server;
     server_name crash-stats;
 
+    root /data/socorro/webapp-django;
```
peterbe: Note the Best Practice of using no trailing spaces.

puppet/modules/socorro/files/etc_nginx/conf_d/socorro-webapp.conf
```diff
@@ -9,13 +11,21 @@ server {
     # crash-stats needs to accept debug symbol zips
     client_max_body_size 1g;
 
-    location / {
+    location /static/CACHE {
+        add_header   Cache-Control public;
+        expires      max;
+        access_log   off;
+        break;
```
peterbe: This `break` is there to tell it to not bother any other `location` stanzas. It'll still proceed procedurally down to line 23.

puppet/modules/socorro/files/etc_nginx/conf_d/socorro-webapp.conf
```diff
@@ -9,13 +11,21 @@ server {
...
+    }
+
```
peterbe: What we might consider, which I have on my personal projects, is that static URLs that don't have md5'ish names (e.g. `/static/browser.js`) we'd just set...
```
add_header Cache-Control public;
expires 1h;
access_log off;
```
at least.

jdotpz:
I'll manually update stage to validate, and then r+ and merge.

jdotpz:
r+, waiting on tests and then we'll merge.

PR #212, opened 2015-08-20
fixes bug 1196781 - Gzip static assets

peterbe:
@jdotpz r?


peterbe:
This assumes that the HttpGzipModule is installed but I find it unlikely that it's not by default. All my debian experience with Nginx (and on OSX too) have always had it installed by default.

PR #213, opened 2015-08-21
Fix bug 1196931 - Upsize analysis server

jdotpz:


peterbe:
Is this not needed any more? 
I see that https://bugzilla.mozilla.org/show_bug.cgi?id=1196931 is now resolved. I'm guessing the fix in that bug was manually changing it in the console.

rhelmer:
@peterbe we want the terraform config to match reality, even if it was manually changed ultimately.

peterbe:
@rhelmer so, is it r+ ?

Sorry for not knowing what I'm talking about. I'm just trying to avoid losing mometum on crash-analysis because it's a critical but fragile beast.

lonnen:
This is good to merge.

PR #215, opened 2015-10-07
Fixes bug 1212485, forces crash dir to be owned properly

jdotpz:


twobraids:
r+ with the assumption that you know what you're doing

PR #217, opened 2015-10-08
Add explicitly named user socorro

jdotpz:


rhelmer:
I think the problem here is in the RPM install, if this gets things rolling as a temporary measure that's fine. If puppet creates the user before installing the RPM I think the rpm install will fail, though.

jdotpz:
That is likely right.  However, we're trying to hurry and get this patched to unplug the pipe, so we're gonna try it.

PR #221, opened 2015-10-14
We care about our servers, so much.

jdotpz:
So we want our deploy to sound the red alert if it can't get healthy nodes into the load balancers.


peterbe:
code looks good but what's with the stage-collector failing?

lonnen:
looks spurious

PR #231, opened 2016-02-05
Ensure logs are rotated on longer running nodes

jdotpz:


puppet/modules/socorro/files/etc_nginx/conf_d/backup
```diff
@@ -0,0 +1,68 @@
+server {
```
peterbe: What's this file?

puppet/modules/socorro/files/etc_logrotate/syslog
```diff
@@ -0,0 +1,18 @@
...
+    missingok
+    rotate 7
+    compress
+    delaycompress
+    notifempty
+    copytruncate
+    postrotate
+  /bin/kill -HUP `cat /var/run/syslogd.pid 2> /dev/null` 2> /dev/null || true
+    endscript
+}
```
peterbe: I thought we already had logrotate on all things. Do we not? So I'm not sure I understand the change. 
Either way, for the admin nodes we might need to add `/var/log/socorro/crontabber.log`

jdotpz: @peterbe we have logrotate, but this is nicer.  It was:

```/var/log/cron
/var/log/maillog
/var/log/messages
/var/log/secure
/var/log/spooler
{
    sharedscripts
    postrotate
	/bin/kill -HUP `cat /var/run/syslogd.pid 2> /dev/null` 2> /dev/null || true
    endscript
}```

This will clean files > 7 days, and also compress ones older than 2 days.

Comment (peterbe) [inline puppet/modules/socorro/files/etc_logrotate/syslog]: Then what about `/var/log/socorro/`?

Comment (jdotpz) [inline puppet/modules/socorro/files/etc_logrotate/syslog]: No files ever seem to be written there, so it shouldn't be necessary to add that dir.

Comment (peterbe) [inline puppet/modules/socorro/files/etc_logrotate/syslog]: > No files ever seem to be written there, so it shouldn't be necessary to add that dir.

On the admin nodes, we do write to `/var/log/socorro/crontabber.log`. I tail -f that file often.

**Comment by peterbe** [comment]
I guess that's good. And it'll make it easier to check the crontabber logs :)
```
[centos@prod-admin-i-ecb76837 ~]$ ls -lh /var/log/socorro/crontabber.log
-rw-r--r-- 1 socorro socorro 3.6G Feb 21 18:30 /var/log/socorro/crontabber.log
```
I don't know what the travis failures mean.

peterbe:
:palm_tree: 
r+ if you can fix the travis thing. Or explain why it's supposed to not work :)

lonnen:
Terraform (and all hashicorp binaries) are now hosted on their own site and not bintray. https://releases.hashicorp.com/terraform/

We will need to change our references accordingly

lonnen:
This should fix it. Rerun after this lands: https://github.com/mozilla/socorro-infra/pull/232

lonnen:
:+1: good to land

lonnen:
... or my page was stale

PR #232, opened 2016-03-15
update terraform location

lonnen:
Terraform binaries have moved from bintray to releases.hashicorp.com and the URL structure has changed slightly. This should fix things up for us.


jdotpz:
arrrr+

PR #247, opened 2016-08-10
Fix the Packer phase link

mortonfox:


phrawzty:
Hello, and thank you for the PR.

This does fix the link if accessed through Github, but oddly enough, you've discovered a doc bug in that this wasn't supposed to be a link in the first place. :smile: This should be a fixed-width quote, as in:

> This case deals with the Packer (`packer/`) phase. There is no reasonable default.

If you'd like to amend your PR in this fashion, I'd be happy to merge it (otherwise I'll fix it myself, heh).

Thanks again!

mortonfox:
Hi,
I have amended the commit as requested.

phrawzty:
Great! Thanks again for the PR.

PR #248, opened 2016-08-11
Upsize staging node to prevent resource exhaustion

jdotpz:


peterbe:
If this works out, let's close https://bugzilla.mozilla.org/show_bug.cgi?id=1281544

willkg:
From IRC:

```
<willkg> jp: it died on "puppet-lint: invalid option: --no-80chars-check". is puppet-lint pinned (or whatever the gem equivalent is)?
<willkg> jp: where "it" is travis-ci.
<lonnen> R+ on content
<lonnen> That lint check is concerning but I wouldn't let it hold us up
```

PR #254, opened 2016-08-23
Remove  as we are now setting it in the job

jdotpz:


willkg:
I verified we're setting the `PYTHON` environment variable in the job. Even if we weren't, the scripts that get run by `make package` will create a `PYTHON` if it doesn't exist with default value `python`.

So, this looks good to me. r+!

PR #259, opened 2016-10-20
bug 1311410 - up nginx file upload size limit

peterbe:


peterbe:
I checked with @luser and he said this is the right move. It gives us breathing room and it's really non-trivial to attack the problem of why the files are so large in the first place.

jdotpz:
LGTM as soon as travis-ci is ready to go.
I've updated the /home/centos/socorro-infra repo on the buildbox, and I am going to trigger a new staging build with it.

PR #260, opened 2016-11-14
bug 1136371 - enable HSTS for crash-stats

peterbe:
Much better place.

### DIFF
--- a/puppet/modules/socorro/files/etc_nginx/conf_d/socorro-webapp.conf
+++ b/puppet/modules/socorro/files/etc_nginx/conf_d/socorro-webapp.conf
@@ -39,6 +39,8 @@ server {
         application/javascript
         text/x-js;
 
+    # HSTS for 1 year
+    add_header Strict-Transport-Security "max-age=31536000";
 
     # crash-stats needs to accept debug symbol zips
     client_max_body_size 2g;
--- a/puppet/modules/socorro/files/etc_nginx/conf_d/socorro-webapp.conf
+++ b/puppet/modules/socorro/files/etc_nginx/conf_d/socorro-webapp.conf
@@ -39,8 +39,11 @@ server {
         application/javascript
         text/x-js;
 
-    # HSTS for 1 year
-    add_header Strict-Transport-Security "max-age=31536000";
+    # HSTS for 1 day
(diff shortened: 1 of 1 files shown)


willkg:
I don't really know much about nginx configuration. So while I could review this, I don't really know good from bad or how this affects our environment.

peterbe:
Do you want to defer to Miles? 
I'm pretty confident this change is good though. It's what I use on a personal project.

willkg:
One thing is that @milescrabil (wow--that's irritating, why isn't he listed in the github autocomplete) or @jdotpz should review this because they're the ops side.

While you might use it on your personal projects, that doesn't cover how this could affect the Socorro environment and users of Socorro.

The bedrock folks started with 1 day and watched for errors (I'm not sure what errors they watched for) and ratcheted it up from there after they were confident. Maybe that's a better path?

peterbe:
@milescrabil is not in the autocomplete because he hasn't committed or participated in the git repo. 

We did a similar thing with air mozilla. First one 5 minutes, for a week then changed it up to on year. 

Also, I'm not sure this is OPs related. There are other webdev related things in there like [cache settings for example](https://github.com/peterbe/socorro-infra/commit/1199debc3b530f2bc25cf5f8929873a91f7dad7b)

In airmozilla, I did the security header stuff via Django code because I don't have access to the Apache config and it's inconvenient to communicate with them via webops bugzilla bugs.

peterbe:
Updated. Let's see if there are any sites that NEED to access crash-stats.allizom.org or crash-stats.mozilla.org over HTTP.

milescrabill:
@jvehent thoughts?

jvehent:
This looks good to me. I usually don't use double-quotes in nginx values, so maybe just use `add_header Strict-Transport-Security max-age=86400;`

peterbe:
[nginx-boilerplate](https://github.com/nginx-boilerplate/nginx-boilerplate/blob/master/boilerplate/enable/ssl.conf#L13) uses single quotes.

milescrabill:
lgtm

peterbe:
@milescrabill Feel free to merge this PR if you approve.

PR #261, opened 2016-11-18
Get proper Packer return code using PIPESTATUS

milescrabill:
Should help debugging Packer failures in AMI creation.

@phrawzty


phrawzty:
I'm not entirely sure why I'm @'d here - @jdotpz wrote this script.  :+1: from me though, fwiw.

PR #262, opened 2016-11-29
fixes bug 1136371 - HSTS for a year

peterbe:
@milescrabill I suggest we let it simmer in stage and prod with the current 1 day expiry and when we know it's not breaking anything then we merge this.


lonnen:
see #278

PR #263, opened 2016-11-30
webapp: increase root block device to 20gb per https://bugzilla.mozilla.org/show_bug.cgi?id=1321323

milescrabill:
@relud r?


relud:
assuming you know their ami supports this, r+

milescrabill:
It looks like it does.

PR #228, opened 2016-01-15
Change the health check to use http looking at a json file to ensure …

jdotpz:
…we don't exercise the db and cache so heavily

PR #265, opened 2016-11-30
add proper HTTP healthcheck to webapp

milescrabill:
@peterbe r?

Closed https://github.com/mozilla/socorro-infra/pull/228 in favor of this PR.

### DIFF
--- a/terraform/webapp/main.tf
+++ b/terraform/webapp/main.tf
@@ -144,7 +144,7 @@ resource "aws_elb" "elb-socorroweb" {
       healthy_threshold = 2
       unhealthy_threshold = 2
       timeout = 3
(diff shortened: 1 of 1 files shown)


peterbe:
I honestly don't know how to review this. I don't feel confident enough around ELB stuff. But if you think this is the right thing to do (i.e. talking to nginx AND django (via uwsgi)) then the URL looks right at least. Merge if you feel confident.

milescrabill:
I'm confident this is the correct syntax for Terraform. @relud r?
You can verify that this endpoint works with a simple: curl -i -L 'crash-stats.mozilla.com/monitoring/healthcheck/?elb=true'.

PR #266, opened 2016-12-01
webapp nginx: add proxy headers, make ELB use HTTPS health check

milescrabill:
@peterbe @relud r?

### DIFF
--- a/terraform/webapp/main.tf
+++ b/terraform/webapp/main.tf
@@ -94,6 +94,14 @@ resource "aws_security_group" "ec2-socorroweb-sg" {
             "${var.elb_master_web_sg_id}"
         ]
     }
+    ingress {
+        from_port = 443
+        to_port = 443
+        protocol = "tcp"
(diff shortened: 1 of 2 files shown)


terraform/webapp/main.tf
```diff
@@ -144,7 +144,7 @@ resource "aws_elb" "elb-socorroweb" {
       healthy_threshold = 2
       unhealthy_threshold = 2
       timeout = 3
-      target = "HTTP:80/monitoring/healthcheck/?elb=true"
+      target = "HTTPS:80/monitoring/healthcheck/?elb=true"
```
relud: 1. didn't you change the port to 443?
1. shouldn't this still be http, because there's no ssl between the elb and the app?

milescrabill: Yay, this is why reviews. Changed HTTP -> HTTPS instead of 80 -> 443 😞

PR #268, opened 2016-12-02
webapp: fix ELB health check by using nginx to set HTTP_HOST

milescrabill:
@relud r?

The issue here is that uwsgi / django were not properly allowing a connection to the health check endpoint when HTTP_HOST was something other than expected. This caused local curls (host localhost) and AWS ELB Health Checks (host the IP of the instance) to fail.

### DIFF
--- a/puppet/modules/socorro/files/etc_nginx/conf_d/socorro-webapp.conf
+++ b/puppet/modules/socorro/files/etc_nginx/conf_d/socorro-webapp.conf
@@ -72,6 +72,14 @@ server {
     proxy_redirect off;
     proxy_http_version 1.1;
 
+    location /monitoring/healthcheck/ {
+        # sets HTTP_HOST for AWS ELB Health Check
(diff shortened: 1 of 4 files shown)


puppet/modules/socorro/files/etc_nginx/conf_d/socorro-webapp.conf
```diff
@@ -72,6 +72,14 @@ server {
     proxy_redirect off;
     proxy_http_version 1.1;
 
+    location /monitoring/healthcheck/ {
+        # sets HTTP_HOST for AWS ELB Health Check
+        uwsgi_param HTTP_HOST 'crash-stats.allizom.org';
```
relud: is it always allizom?

milescrabill: I was hoping that it would be ok, but it seems the webapp is finicky with the hostname.

milescrabill:
Merging for justice.

PR #271, opened 2016-12-02
remove stage-submitter from stage_socorro_master.list, add it to roles in wrapper.sh

milescrabill:
These changes were unmerged sitting on the buildbox. I figured I would preserve them.

The implications, best as I can tell:
Removing stage-submitter from stage_socorro_master.list means that stage-submitter will no longer be touched by deploy-socorro.sh. It is my impression that this was already the case, because in identify_role.sh it was listed as "submitter-stage" and not "stage-submitter". This is bizarre, the effect being that stage-submitter was _probably_ not touched. The last time the instance was modified was Oct 13th, 2016, which implies that it probably _was_ touched.

Ignoring this.

@lonnen @willkg @peterbe thoughts? I don't know much about the stage-submitter.

### DIFF
--- a/bin/lib/identify_role.sh
+++ b/bin/lib/identify_role.sh
@@ -31,7 +31,7 @@ function identify_role() {
             SCALEVARIABLE="processor_num.stage"
             ;;
 
-        submitter-stage )
+        stage-submitter )
             AUTOSCALENAME="as-stage-submitter"
             ELBNAME="elb-stage-socorroweb"
             TERRAFORMNAME="stagesubmitter"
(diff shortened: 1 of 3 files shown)


willkg:
The stage submitter doesn't update when we do deploys. It's a long-running instance that I took over a while back--possibly in October. Any time I do work on the box to update it, change configurations, etc, I email the ops folks in charge (i.e. @milescrabill, @jp).

I haven't figured out what the future of the stage submitter is. Clearly things change when we switch to Antenna. I don't know how, yet.

Given that, I'm inclined to just leave the stage-submitter as is: manually managed long-running instance banished to a tower overgrown with brambles and poison ivy in a dark brooding forest inhabited only by spiders.

milescrabill:
I'm thinking that this is needed. If stage-submitter stays in stage_socorro_master.list it should be attempted to be deployed.

@relud r?

PR #276, opened 2016-12-07
text/html from gzip config in webapp nginx

milescrabill:
@relud r?

Turns out text/html is implicit and declaring it is a duplicate


lonnen:
💯

lonnen:
:micronesia:

PR #281, opened 2016-12-13
use python2-pip instead of python-pip, which is no longer a package

milescrabill:
I'm just going to merge this.


lonnen:
retro r+

PR #282, opened 2017-01-03
Use Terraform to create updated SSL Cipher Policy for collector oldssl

milescrabill:
@relud r?


milescrabill:
Merging to test in stage.

PR #283, opened 2017-01-04
use terraform plan with -refresh=false

milescrabill:
just testing for now


milescrabill:
Per the meeting on 01-04-17, I will add the AWS creds to TravisCI so that this works.

PR #286, opened 2017-01-04
Fix a case where multiple AMIs are found for a given commit hash

milescrabill:
@relud or @lonnen r?

### DIFF
--- a/bin/deploy-socorro.sh
+++ b/bin/deploy-socorro.sh
@@ -279,17 +279,16 @@ function terminate_instances() {
 
 function find_ami() {
     PROGSTEP="Find correct prod AMI"
-    echo "`date` -- Attempting to locate AMI tagged with appsha of ${GITCOMMITHASH}"
+    echo "`date` -- Attempting to locate AMI tagged with appsha ${GITCOMMITHASH}"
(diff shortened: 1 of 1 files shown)


bin/deploy-socorro.sh
```diff
@@ -286,7 +286,7 @@ function find_ami() {
     echo "`date` -- Attempting to locate AMI tagged with appsha of ${GITCOMMITHASH}"
         NEWAMI=$(aws ec2 describe-images \
                  --filters Name=tag:apphash,Values=${GITCOMMITHASH} \
-                 --output text --query 'Images[*].ImageId')
+                 --output text --query 'Images[*].ImageId' | awk '{print $1;}')
```
relud: random code golf:

- `'Images[]'` does the same thing as `Images[*]`
- `'Images[].[ImageId]' | head -1` is shorter than awk
- `'Images[].ImageId' | cut -f1` is even shorter

milescrabill: 'Images[0].ImageId' wins

willkg: OMG. I <3 that you all do code golf.

PR #287, opened 2017-01-04
Better error handling for find_ami and apply_ami

milescrabill:
@relud or @lonnen r?


milescrabill:
Best I can tell $GITHUBHASH was only being used for manual runs of deploy-socorro.sh, where it was manually set as an env var, and the script was manually run.

I'm working on a cleaned up, env specific version that will hopefully make manual runs nonexistent or better.

PR #288, opened 2017-01-06
fix deregister_elb_nodes() to properly query ELBs

milescrabill:
@relud r?


milescrabill:
It's pretty likely that this issue is related to the collector alerts that trigger _sometimes_ during deploys. It seems as though proper connection draining was not happening! Instead, instances were terminated without being removed from the ELB. @willkg

PR #289, opened 2017-01-06
add stage deploy script adapted from deploy-socorro.sh to be run via CI

milescrabill:
epic bash script legacy continues

@relud r?

### DIFF
--- a/bin/deploy-stage.sh
+++ b/bin/deploy-stage.sh
@@ -1,0 +1,373 @@
+### this is a minimal script that executes socorro stage build
+### takes on arguments, builds most recent commit to master
+
+### adapted from deploy-socorro
+
+set +u
+set +x
+
+SCRIPT_PATH=`dirname $(realpath $0)`
+SOCORRO_INFRA_PATH="/home/centos/socorro-infra"
+ENVIRONMENT="stage"
+ENDRC=0
+
+STARTLOG=`mktemp`
+ENDLOG=`mktemp`
+
+# Roles / instance types to deploy for stage
+ROLES=`cat ${SCRIPT_PATH}/lib/${ENVIRONMENT}_socorro_master.list`
+
+# imports
+. ${SCRIPT_PATH}/lib/identify_role.sh
+. ${SCRIPT_PATH}/lib/infra_status.sh
+
+get_stage_git_info() {
+    echo "git info for ${PWD}"
+    # short sha of latest commit from master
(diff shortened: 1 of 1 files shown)


bin/deploy-stage.sh
```diff
@@ -0,0 +1,373 @@
...
+
+    HEALTHYHOSTCOUNT=$(aws elb describe-instance-health \
+                       --load-balancer-name ${ELBNAME} \
+                       --output text | awk '{print $5}' | \
+                       grep InService | wc -l | awk '{print $1}')
+    RC=$?; error_check
+
+    CURRENTHEALTH="${CURRENTHEALTH} $(aws elb describe-instance-health \
+                   --load-balancer-name ${ELBNAME} \
+                   --output text | awk '{print $5}')"
```
relud: I think it would be more transparent to use `--query` on the command on line 155 and this one instead of awk'ing out columns. A query for the first command would be: <pre>--query 'length(InstanceStates[?State==`InService`])'</pre>

bin/deploy-stage.sh
```diff
@@ -0,0 +1,373 @@
...
+    # For each of our apps, we want to use terraform to apply the new base AMI we've just created
+    for ROLEENVNAME in $ROLES; do
+            # Get AS group name for each ROLE
+            echo "`date` -- Checking role for ${ROLEENVNAME}"
+            identify_role ${ROLEENVNAME}
+            infra_report ${AUTOSCALENAME} >> ${STARTLOG}
+            ASCAPACITY=$(aws autoscaling describe-auto-scaling-groups \
+                         --auto-scaling-group-names ${AUTOSCALENAME} \
+                         --output text \
+                         --query 'AutoScalingGroups[*].DesiredCapacity')
```
relud: again maybe 0 instead of *?

bin/deploy-stage.sh
```diff
@@ -0,0 +1,373 @@
...
+function terminate_instances_all() {
+    STEP="Terminating instances"
+    for ROLEENVNAME in $ROLES; do
+        # First, we halve the number of minimum size for each group
+        echo "`date` -- Setting min size for ${ROLEENVNAME}"
+        identify_role ${ROLEENVNAME}
+        ASCAPACITY=$(aws autoscaling describe-auto-scaling-groups \
+                     --auto-scaling-group-names ${AUTOSCALENAME} \
+                     --output text \
+                     --query 'AutoScalingGroups[*].DesiredCapacity')
```
relud: again maybe 0 instead of *?

PR #291, opened 2017-01-06
Fix up deploy-stage.sh

milescrabill:
clean up after ourselves, fix some small bugginess, debug mode in bash

@relud r?

### DIFF
--- a/bin/deploy-stage.sh
+++ b/bin/deploy-stage.sh
@@ -4,7 +4,7 @@
 
 ### adapted from deploy-socorro
 
-set -u
+set +u
 set +x
 
 SKIP_TO_DEPLOYMENT="false"
@@ -14,6 +14,11 @@ if [[ -n $1 ]]; then
     RPM="RPM has no value, RPM building was skipped."
     AMI_NAME="AMI_NAME has no value, RPM building was skipped."
     SKIP_TO_DEPLOYMENT="true"
+fi
+
+FORCE_REBUILD="false"
+if [[ -n $2 ]]; then
+    FORCE_REBUILD="rebuild"
 fi
 
 SCRIPT_PATH=`dirname $(realpath $0)`
@@ -70,6 +75,7 @@ clone_repo() {
     RC=$?; error_check
 
     cd $TMP_DIR/socorro
+    STEP="[clone_repo] Acquiring git metadata"
     # print socorro commit info
     get_stage_git_info
 }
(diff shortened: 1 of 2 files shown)


bin/deploy-stage.sh
```diff
@@ -92,16 +96,16 @@ function create_ami() {
     /usr/bin/packer build -color=false $SOCORRO_INFRA_PATH/packer/socorro_base.json | tee ${TMP_PACKER_LOG}
     RC=${PIPESTATUS[0]}; error_check
     # Assign the sparkly new AMI id to a variable
-    AMI_ID=`cat ${TMP_PACKER_LOG} | grep us-west-2 | grep ami | awk '{print $2}'`
+    AMI_ID=`grep ${TMP_PACKER_LOG} 'us-west-2: ami-*' | cut -f 2 -d ' '`
```
relud: doesn't grep require the expression before the filename? also do you mean `ami-.*` because regex not glob?

bin/deploy-stage.sh
```diff
@@ -371,4 +375,7 @@ echo "==========  ENDING STATE  ==========="
 cat ${ENDLOG}
 rm ${STARTLOG}
 rm ${ENDLOG}
+if [[ $TMP_DIR =~ "^/tmp/*" ]]; then
```
relud: again `.*` because regex?

bin/deploy-stage.sh
milescrabill: ouch, yeah the order is wrong

milescrabill:
A lot has happened, this script is now quite a bit more complicated. I've added a README that explains the high level workings of the script.

It's also maybe done?

bin/deploy-stage.sh
```diff
@@ -1,374 +1,504 @@
...
+### optional argument $2: "rebuild" to force a rebuild of the SHA (rpm + ami)
 
 ### adapted from deploy-socorro
 
 set +u
 set +x
 
-SCRIPT_PATH=`dirname $(realpath $0)`
+format_logs() {
+    # requires figlet installed
```
willkg: It's probably worth adding a section to the top that verifies all the required things are installed. I was doing this in the Antenna system tests:

https://github.com/mozilla/antenna/blob/master/tests/systemtest/run_tests.sh#L37

milescrabill: I've updated the script to check dependencies. I attempted something similar to your fancy function, but I figured it was an exercise in putting lipstick on a pig in this case and reverted away.

bin/deploy-stage.sh
```diff
@@ -1,374 +1,561 @@
...
+. "${SCRIPT_PATH}/lib/identify_role.sh"
+. "${SCRIPT_PATH}/lib/infra_status.sh"
+
+# for postgres and python
+PATH="${PATH}:/usr/pgsql-9.3/bin:/usr/local/bin/"
+echo "PATH: ${PATH}"
+
+check_for_dependencies() {
+    STEP="[check_for_dependencies] checking if aws in PATH"
+    aws help > /dev/null
```
milescrabill: No particular reason, heh. I almost did use which, but didn't.

PR #292, opened 2017-01-11
replace evil awk/sed/head usage with jq

milescrabill:
@relud r?


milescrabill:
Let it be known that the tests failing is meaningless with regards to this PR.

PR #295, opened 2017-01-17
Fix PATH for dependency checking, fix find_ami no ami found case

milescrabill:
@relud r?

### DIFF
--- a/bin/deploy-stage.sh
+++ b/bin/deploy-stage.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 ### this executes socorro stage build and can run without arguments
 ### by default builds most recent commit to master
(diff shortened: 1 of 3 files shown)


bin/deploy-stage.sh
```diff
@@ -97,27 +99,30 @@ check_for_dependencies() {
     RC=$?; error_check
 
     STEP="[check_for_dependencies] checking if terraform in PATH"
-    TERRAFORM_VERSION=$(terraform version | grep -o 'Terraform v[0-9]*\.[0-9]*\.[0-9]*')
+    TERRAFORM=$(which terraform)
```
relud: this which thing is unnecessary, because you fixed the path.

milescrabill:
Updated this PR to account for the now merged deploy-prod.sh and deploy_functions.sh. And yes, those names are consistent with the previous naming scheme. Ha.

PR #302, opened 2017-03-20
deploy-prod: use nonrelative tag by default

milescrabill:
@relud r?


relud:
what do tags look like that this is cutting?

milescrabill:
Goes from something like `263-32-gd39c56670` (tag 32 commits behind master) and cuts to `263`.

PR #305, opened 2017-04-24
Bug 1358577: Reverse order of crashmover crashstorage classes

willkg:
It's better to have the BotoS3CrashStorage *before* the RabbitMQCrashStorage so
as to reduce the race condition where the processor goes to process a crash that
isn't on S3, yet.

This switches the order of `destination.storage_classes` and then swaps the `storage0` and `storage1` configurations.


willkg:
@peterbe Can you eyeball this? It's not used in Socorro infrastructure--it's just an example config file.

Travis is failing on linting of puppet scripts which is unaffected by this change.

willkg:
It's probably fine. I'm going to land it.

PR #309, opened 2017-07-31
Make HTTP redirect use servername

claudijd:
This simply sets the servername and uses it in it's redirect behavior.

Please verify this change, as I did not test it.


claudijd:
Based on the build failure log, these failures look to be unrelated to the changes I made (ie. complaints about "=>" indentation in puppet//modules/socorro/manifests/role/webapp.pp).  If there is something I did wrong or I need to run the nginx config through a linter first, please let me know.

milescrabill:
@claudijd yeah, due to misc. sadness we ignore CI on this repo. I'll disable it. This PR lgtm.

PR #310, opened 2017-08-15
socorro-infra: use cloudops nodesource mirror to override EPEL node

milescrabill:
@relud r?


milescrabill:
This was necessary because of https://bugzilla.redhat.com/show_bug.cgi?id=1481470

PR #317, opened 2018-02-16
packer: delete volumes on termination

milescrabill:
I should have done this so so so long ago.
See notes here: https://www.packer.io/docs/builders/amazon-ebs.html#ami_block_device_mappings
Basically, this is the cause of our volume woes: https://bugzilla.mozilla.org/show_bug.cgi?id=1298049

### DIFF
--- a/packer/socorro_base.json
+++ b/packer/socorro_base.json
@@ -13,7 +13,11 @@
             "source_ami": "ami-6935e909",
             "instance_type": "m3.medium",
             "ssh_username": "centos",
-            "ami_name": "socorro_centos7__{{timestamp}}"
+            "ami_name": "socorro_centos7__{{timestamp}}",
+            "launch_block_device_mappings": [{
+                "device_name": "/dev/sda1",
+                "delete_on_termination": true
+            }]
         }
     ],
     "provisioners": [


milescrabill:
I believe the problem is some of #1 and some of #2. There are leftover volumes created by Packer and simply not cleaned up. There are also snapshots that correspond to AMIs created by Packer, and neither the snapshots nor the AMIs are cleaned up, either.

In a test run of Packer with this change, the volume id for the Packer instance was vol-006972c96b7d6a57b, and after the AMI was created a search for that id in snapshots yields snap-0cb4474b577b515ae, with description "Created by CreateImage(i-0e104a244efc07ec6) for ami-08c84570 from vol-006972c96b7d6a57b." The volume was deleted, but the snapshot wasn't, because when an instance is created from the AMI the snapshot is used to create its root volume.

We're not filled up on snapshots, though: there were 8x as many volumes as snapshots prior to me deleting a bunch of volumes. Further, quite a few volumes reference the same snapshot - so maybe we have some of problem #2 as well. To me, this indicates that we created volumes from the AMI and _those_ volumes were never deleted. They reference the same snapshot because they were created from the AMI that used that snapshot. That's my guess. The Packer docs indicate that delete_on_termination defaults to true for these volumes - but maybe that wasn't the case when our version was released.

In response to your points about why this change shouldn't work: in testing, specifying the `launch_block_device_mappings` the way I did sets /dev/sda1 as the root and only block device attached to the instance. Looking at existing AMIs, they all use a root device at /dev/sda1.

The Packer version on Leeroy is 0.7.5. Upgrading required some syntactic changes that I never went through with, see [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1322566).

sciurus:
My mullings about why this _shouldn't_  works were based on my interpretation of the Packer documentation for the launch_block_device_mappings options. I'll totally believe it really _does_ work, since you've tested it and https://github.com/hashicorp/packer/issues/1783 recommends it. I found https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/block-device-mapping-concepts.html which suggest that sda1 instead of sda is just a quirk of the root device, and that if you specify a mapping for sda1 it will modify some parts of the default mapping. That makes what packer is doing make more sense.

